PR #17, opened 2023-09-15
added spinner next to progressBar

driverag22:
spinner added next to the progressBar. The code for the spinner can be found in progressBar.ts, with two simple functions to start and stop the spinner and a few lines to add the html.

spiner.css defines the style.

The progressBar width was slightly reduced, and the background color was changed to have better contrast and better show the progress (it is easier to tell what 100% of the progessBar is this way). This was needed because with the addition of the spinner it was hard to tell where the progressBar ends.

### DIFF
--- a/editor/src/kroqed-editor/editor.ts
+++ b/editor/src/kroqed-editor/editor.ts
@@ -96,6 +96,7 @@ export class Editor {
 			// Hack to forcefully remove the 'old' menubar
 			document.querySelector(".menubar")?.remove();
 			document.querySelector(".progress-bar")?.remove();
+			document.querySelector(".spinner-container")?.remove();
 			this._view.dom.remove();
 		}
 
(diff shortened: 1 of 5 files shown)


jim-portegies:
@driverag22 After zooming in a bit or zooming out a bit, the spinner is not at the same height as the progress bar anymore. Can you look into this?

driverag22:
> @driverag22 After zooming in a bit or zooming out a bit, the spinner is not at the same height as the progress bar anymore. Can you look into this?

This is now fixed, I just had to change the spinner height to be relative to the font-size using "rem" as a unit of measurement.

When zooming we also get that at some point the progressBar and spinner overlap. Is this fine? If not I could add some more padding (and hence reduce the width of both) so this does not happen with "normal" levels of zoom, or look into other options.

jim-portegies:
@driverag22 if you open a document that takes some time to evaluate, and you immediately scroll down a bit, I feel the spinner is still not at the right place.

driverag22:
> @driverag22 if you open a document that takes some time to evaluate, and you immediately scroll down a bit, I feel the spinner is still not at the right place.

This is also now fixed. Due to the css position the spinner had, when scrolling down it's top margin would collide with the top part of the editor and the spinner would indeed move down.

Regarding the width of the progressBar, right now both the progressBar and spinner fit snugly in the same line, but if you resize the window to be smaller at some point the spinner gets pushed down below the progressBar. There is no real work around this, but I can resize the progressBar to be even smaller so this only happens if the window is resized to be very small. If this is not a problem, I can also make the progressBar a bit bigger so it fills out the space a bit more.

PR #35, opened 2023-10-21
Added activeLineIndicator for focused code blocks

driverag22:
Enabled the activeLineIndicator and made the indicator be transparent for unfocused code blocks so it only "shows up" for the focused code blocks.
The color of the activeLine can be changed if needed.


jim-portegies:
Very nice!

PR #38, opened 2023-10-27
Client: Use the beginning of the sentence when executing commands.

DikieDick:
The beginning of the current sentence (ie. the sentence that the cursor is currently in) refers to the end position of the previous sentence, since we do not have enough information to actually find the beginning.

This PR changes the working of the `commandExecutor` such that it now uses the end of the previous sentence when executing commands, instead of the end.

### DIFF
--- a/src/lsp-client/client.ts
+++ b/src/lsp-client/client.ts
@@ -158,6 +158,11 @@ export function CoqLspClient<T extends ClientConstructor>(Base: T) {
             return this.start();
         }
 
+        getBeginningOfCurrentSentence(): Position | undefined {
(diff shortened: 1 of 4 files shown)


DikieDick:
The `Help.` command now executes the help for the next line when the cursor is positioned right after a `.` character.

DikieDick:
@jim-portegies Could you test this? 
Do I need to filter the error message we get when we execute the command at the end of a finished proof?

jim-portegies:
Looks good! We currently don't need to filter because it is a bit too complicated because it is very context dependent. Later it may be a good idea.

PR #40, opened 2023-11-02
Added coq-syntax option to change tactics which show up for autocompletion

driverag22:
Added new tacticsCoq.json file with the normal coq-syntax tactics.

Also added a setting to change coq-syntax. The information for the selected syntax is passed to editor.ts, which initializes the tactics.ts singleton accordingly.

Laslty, implemented singleton method in tactics.ts to use normal or coqTactics based on how the singleton is initialized by editor.ts Editor.ts "knows" what to initialize the singleton to based on the setting selected by the user. Moreover, the singleton can be "overriden" (i.e., can be initialized more than once) such that if the setting changes, the singleton will change accordingly.

coqTacticsData.json was also created and tactics were added, but with no description and very basic examples. Showing the coq tactics under the tactics menu is not yet implemented.


jim-portegies:
@driverag22 If I open the tactics panel, I still see the Waterproof tactics. Was this intended?

driverag22:
> @driverag22 If I open the tactics panel, I still see the Waterproof tactics. Was this intended?

Yeah, since we weren't sure how we wanted it I left it like this for now. We can freeze the PR for now and I can implemented that sometime at the start of Q2 if needed.

jim-portegies (changes requested):

jim-portegies (approved):

PR #50, opened 2023-11-17
Fixed error that showed up due to <tr> being a child of <table> directly with no <tbody> in between

driverag22:
As far as I can tell, the error only showed up in the console log and did not affect user experience or anything else in any way.
The way to recreate the error is simply opening the Expand definition panel and the error message will print out. This fix removes it.


jim-portegies (approved):

jim-portegies:
Looks good!

PR #58, opened 2023-11-27
added check for periods in expand definition

driverag22:
Whenever the given input ends with a period followed by 0 or more spaces, we disregard them. This only happens if the sentence ends with that combination, so a period in the middle of the sentence is not removed.


jim-portegies (approved):
Looks good!

PR #59, opened 2023-11-27
Add message on illegal file name

XyntaxCS:
This is the same as before. So I included this pop-up message when opening a file. Please check if you agree with the text in the file (I found it hard to encode " ", a space character). Furthermore, the file does not open right now if it contains one of these characters. Is this a good or a bad thing?


jim-portegies (changes requested):
I've changed the error message, please see if you agree.

Ideally we show the "more aggressive" popup, which would include a button to immediately rename the file. Would this be possible?

jim-portegies (approved):

jim-portegies:
There are still some errors showing up in the Console, let's check these later...

PR #65, opened 2024-01-12
Split expand definition, help and search buttons into three, and added check for a dot/empty lines when using Search/Execute views

driverag22:


driverag22:
Saw some errors popping up, will make sure that everything is fine with the PR before opening it again.

PR #67, opened 2024-01-17
Split expandDef, Help and Search button into three, added fix to remove empty lines/ending dot for input for Search and Execute, fixed right click->help not loading directly

driverag22:


driverag22:
The problem has now been fixed for the expand def. and search panels, although it can still happen with the help panel.

The following steps can be taken to recreate this:
  - Open vscode with the plugin loaded and a file opened from before (make sure not to click anywhere on the file)
  - Open a help webview and click help.
  - It should "bug out" and open a new Help webview.
On main the same steps can be taken with the exp. def. panel and search panels to get the same result.

This is related to the right click->help functionality, and I'm still working on it, with the aim to find a fix that fixes this and fixes right click->help not automatically loading the help response.

I believe that we can merge this for now since it is already less buggy than main, but we can also wait until I figure out a fix for the help panel.

jim-portegies:
@driverag22 There seems to be a bug with right click help. For me it shows up if I have two files open, for instance one other file and the tutorial. If I right-click Help after "We conclude that (3 = 3)." in the tutorial, Help is inserted at a strange place.

driverag22:
@jim-portegies What other file did you have open when this happened, and where do you see Help being inserted?
 I'm trying to recreate the bug but it seems like "Help." is being inserted correctly after the "We conclude that (3=3).".

jim-portegies:
@driverag22  I had the real numbers exercises open, and first performed a Help-query there

driverag22:
@jim-portegies I think I was able to recreate it. The Help got inserted as follows
image
breaking the document.
I'll look into this.

jim-portegies:
The error was already present in the version on main, so merging.

PR #72, opened 2024-03-20
Fix goal panel showing too much information

XyntaxCS:
Removed entire description of subproofs being displayed. Furthermore, ensure the goal panel is always scrolled to the top.


XyntaxCS:
need updates

PR #76, opened 2024-05-24
Changed background color for selected text for more visibility

driverag22:
Some light themes (e.g. solarized light) have a background color for selected text that has very low visibility. This change fixes that and uses a background color that has much better visibility/contrast.


jim-portegies (approved):
Thanks!

PR #77, opened 2024-06-12
Line long errors implementation using coq-lsp diagnostics

driverag22:
Implementation of line long errors with new coq-lsp diagnostics.
Adds two new waterproof settings:
  - `send_diags_extra_data`, which is a coq-lsp required setting, it has to be set to true to get line-long ranges in the diagnostics (and is currently set to true by default)
  - `detailedErrorsMode`, which, when enabled gives detailed errors and when disabled gives line-long errors.

Note that this uses coq-lsp version ~`0.1.9+8.19`~ `0.2.2+8.17`, I have updated the `package.json` and `README.md` accordingly, but if we still don't want to upgrade to the new coq release, we should wait before merging the PR.


DikieDick:
For completeness I repeat here part of the commit message of my last commit.

#### Changes:
  - Implement custom type `WpDiagnostic` that includes the specific coq-lsp fields. Implementing this fixes the typescript compilation issues.
  - Respect changing the `Waterproof.detailedErrorsMode` setting, by listing for a change to this setting in the right place.

#### Note:
If we want line long errors to be the default, we should change the setting `Waterproof.detailedErrorsMode` to be `true` by default.

jim-portegies (approved):
Thanks a lot! There has been a change in `coq-lsp` that needs different handling of the data field of the diagnostic data coming in, thanks to @DikieDick for implementing this. Moreover, with line-long errors there are some situations (for instance beginning of proofs) where you too easily run into what we called 'multiline errors'. I changed the code a bit so that we now support also those multiline errors.

PR #82, opened 2024-09-08
Relax input area strictness

DikieDick:
This change relaxes the strictness of the transaction filtering that we do outside of input areas. Previously only transactions that did not alter the document **and** contained a selection update were allowed (we specifically allowed selection updates to allow selecting code outside of input areas). This has now been changed to allowing all transactions that do not alter the document (ie. satisfy `tr.docChanged == false)`), since hovering over diagnostics does not generate a selection update and we need to allow the diagnostics to be shown.

A nicer approach to filter these transactions might be using `Annotations`, but I could not get this to work using the hover effect of the diagnostics.


jim-portegies:
Great! Thank you!

PR #86, opened 2024-09-24
Make snippets explicit, so no syntax errors appear.

jim-portegies:
The goal is that provides snippets do not give syntax errors, and immediately show what is expected.

For instance, after executing the following lines:
```
Goal forall a : nat, (a > 0) -> True.
Proof.
Take a : nat.
```
the snippet suggestion is ```Assume that (0 = 0).```

### DIFF
--- a/shared/completions/tactics.json
+++ b/shared/completions/tactics.json
@@ -47,7 +47,7 @@
         "label": "Choose (*name_var*) := ((*expression*)).",
         "type": "type",
         "detail": "tactic",
-        "template": "Choose ${x} := (${ 0}).",
+        "template": "Choose ${0:x} := (${1:0}).",
         "description": "You can use this tactic when you need to show that there exists an x such that a certain property holds. You do this by proposing (*expression*) as a choice for x, giving it the name (*name_var*).",
         "boost": 2,
         "example": "Lemma example_choose : \n  there exists y : ℝ,\n    y < 3.\nProof.\nChoose y := (2).\nWe conclude that (& y = 2 < 3).\nQed."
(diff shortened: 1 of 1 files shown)


shared/completions/tactics.json
```diff
@@ -47,7 +47,7 @@
         "label": "Choose (*name_var*) := ((*expression*)).",
         "type": "type",
         "detail": "tactic",
-        "template": "Choose ${(*name_var*)} := (${(*expression*)}).",
+        "template": "Choose ${x} := (${ 0}).",
```
DikieDick, line 50: There is an extra space here.

DikieDick (commented):
Looks good. 

My only remark is that now the tab functionality of jumping to the next 'hole' in the template is less intuitive (I don't know if this functionality is used a lot by students).
Previously the `(* comment *)` placeholders clearly indicated a position where the user could jump to.

DikieDick (dismissed):

DikieDick:
Unrelated to the changes proposed in this PR, but the template for the 'chain of inequalities' inserts an unnecessary closing bracket `)`: [./shared/completions/tactics.json#L77](https://github.dev/impermeable/waterproof-vscode/blob/feat-parsing-templates/shared/completions/tactics.json#L77).

shared/completions/tactics.json
DikieDick: Ah, without the space the `0` character gets interpreted as the order this placeholder should be completed in ([see documentation](https://codemirror.net/docs/ref/#autocomplete.snippet)). The following template works as intented: `"Choose ${0:x} := (${1:0})."`.

jim-portegies: Thanks! That's a very good thing to know.

DikieDick:
@jim-portegies I was mistaken in saying that the closing bracket is unnecessary, somehow I completely overlooked the opening bracket on the same line :)

My original reason for this comment is the following example:
1. Insert `We conclude that (0 = 0).` using the snippet completions.
2. Cursor moves into the `0 = 0` placeholder.
3. Suppose we want to insert a chain of inequalities. We continue with typing `&`, the snippet appears and we insert it into the document.
4. The final result is `We conclude that (& 3 < 5 = 2 + 3 ≤ 7 )).`, which produces a syntax error.

Note: This is works fine using the tactics panel.

jim-portegies:
@DikieDick your point about the snippet insertion was very good. I tried to implement the changes

DikieDick (approved):
Looks good. 

The chain of inequalities works really nicely :)

PR #88, opened 2024-10-05
Show notices as diagnostics by default

jim-portegies:
Best effect with the changes currently in [this PR](https://github.com/impermeable/coq-waterproof/pull/84)


src/lsp-client/clientTypes.ts
```diff
@@ -119,7 +119,7 @@ export interface CoqLspServerConfig {
     client_version: string;
     eager_diagnostics: boolean;
     goal_after_tactic: boolean;
-    show_coq_info_messages: boolean;
+    show_waterproof_info_messages: boolean;
```
DikieDick, line 122:
```suggestion
    show_coq_info_messages: boolean;
```

We can't rename these as they are passed to the `coq-lsp` instance and parsed there. So internally we have to keep the 'coq' name, while for the external/user facing side 'waterproof' is fine.

src/lsp-client/clientTypes.ts
```diff
@@ -139,7 +139,7 @@ export namespace CoqLspServerConfig {
             client_version: client_version,
             eager_diagnostics: wsConfig.eager_diagnostics,
             goal_after_tactic: wsConfig.goal_after_tactic,
-            show_coq_info_messages: wsConfig.show_coq_info_messages,
+            show_waterproof_info_messages: wsConfig.show_waterproof_info_messages,
```
DikieDick, line 142:
```suggestion
            show_coq_info_messages: wsConfig.show_waterproof_info_messages,
```

See comment above.

DikieDick (changes requested):

src/lsp-client/clientTypes.ts
jim-portegies: Thanks! Apparently I didn't know how this worked.

DikieDick (commented):
Looks good to me.

DikieDick (approved):

PR #91, opened 2024-10-10
Insert button for diagnostics

DikieDick:
This PR aims at implementing a new button for diagnostics, with the goal of inserting the diagnostics message at the current cursor position.


DikieDick:
Currently the diagnostic action (labeled "Apply Suggested Action 🔨") just inserts `\n[diagnostic message]` at the cursor position.

We can easily filter and manipulate this based on diagnostic attributes. Examples:
1. **Diagnostic severity**: We should not add this action for `Error` severity message, for example, as it does not make sense to enter the error message itself into the editor.
2. **Pre-process the message text**: Using a general structure in the message like `Add the following line to the proof: "..."`. We can be sure to only insert the actual 'code' part into the editor, without the `Add the following line to the proof` message. In theory we can do even more advanced pre-processing as long as we can communicate the necessary information using the Coq diagnostics.

DikieDick:
I have addressed 2 according to https://github.com/impermeable/coq-waterproof/pull/84, however the actual implementation needs some further polishing.

Furthermore, there is now finally a use for the different styles that apply when a proof is technically correct, but contains a Waterproof warning message:
afbeelding

jim-portegies:
I think the fix with replacing the variable is fun, but perhaps for now it is not strictly necessary anymore (I don't know how stable it is). In general, we could try to provide the fixed statement also on the coq-waterproof side, that might be a more stable solution. This will probably require some work.

The current version does not seem to insert the text anymore in case of another type of warning.

Perhaps some desired functionality (but this needs some experimentation):
* When a line gives an error, "apply fix" could replace the line with a suggestion
* It would be nice if previous/existing indentation is taken into account.

jim-portegies:
Another question: How can we make the look of the diagnostic panel a bit cleaner? In particular:
* Do we still need the 'copy' button?
* Do we need the text 'copy' / 'apply fix' or do the icons suffice? (a hammer is maybe not so suggestive for a fix). Perhaps only 'fix' as text?
* Is there a way to control the alignment of the items in the diagnostics panel?

jim-portegies (approved):
Thanks a lot! The insert functionality is nice.

It would be nice if there would be a bit more feedback after clicking the copy button, but we could make that a different PR.

I have merged main and removed the lezer imports, because I don't think they are necessary at this stage.

PR #94, opened 2024-10-15
Switch to `const enum` in shared files.

DikieDick:
The const enum is compiled away by the typescript compiler. This saves on memory and hopefully also on data that is passed between editor and extension.


DikieDick:
See https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls for possible pitfalls of using `const enum`s.

PR #96, opened 2024-10-22
Make sure messages about different goals are visibly split

pimotte:
Works by adding a black border below the messages about each separate goal. (Even in the case there is only one).

Test by:
Inspecting the goal view after a `Choose :=` and observe that the messages are cleanly split.

### DIFF
--- a/views/styles/goals.css
+++ b/views/styles/goals.css
@@ -52,4 +52,5 @@ p.num-goals + p.aside {
   padding-bottom: 1ex;
   white-space: pre;
   font-size: var(--vscode-editor-font-size);
+  border-style: none none solid none;
 }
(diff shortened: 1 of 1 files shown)


DikieDick (commented):
Looks good to me. See my suggestion in the conversation.

DikieDick:
Just applying the border makes it show up also when there is just one goal present. This looks a bit messy in my opinion. 
Changing the CSS a bit to select only the nodes following an existing goal, using the 'next sibling' selector `+` and adding the border to the top instead of the bottom makes it only show up for multiple goals. 

A little extra fix is needed for a selected goal followed by a subgoal, see bottom code block.

afbeelding

afbeelding

```css
.coq-goal-env {
  padding-top: 1ex;
  padding-bottom: 1ex;
  white-space: pre;
  font-size: var(--vscode-editor-font-size);
}

.coq-goal-env + .coq-goal-env {
  border-style: solid none none none;
}

.box + div>div>div.box {
  border-style: solid none none none;
}
```


I have not tested this extensively or with more than two goals.

pimotte:
Adopted your suggestion:)

Issue #97, opened 2024-10-29
[Bug] Desync in editor


pimotte:
### What happened?
Desync happens when entering a lot of text quickly and then pressing ctrl+z

### How can we reproduce this issue?
Example:
1. Open .mv file in both regular and text view (text view helps for debugging, is not necessary to reproduce)
2. Quickly enter some imports.
3. Quickly press ctrl+z
4. Observe that the text is missed up in the text view.

### What was supposed to happen?
Text content of the file should match what is in the editor

### Additional Information
Please specify the operating system and the version of the Waterproof extension you were using:
#### Operating System: 
Observed on Windows, easier to reproduce under heavy load

#### Waterproof Version: 
- main

--- 

### Extra information or details

Possible avenues to fix this:

- Intercept Ctrl+z and cancel it if edits are inprogress
- Intercept Ctrl+z, add a custom edit event to the queue and fire it programmatically when it's turn in the queue is
- Intercept Ctrl+z, and delegate it to Prosemirror

More general ideas:

- Cancel the queue after resyncing the editor?

Some of these have the disadvantage that it specifically looks at ctrl+z, which is the easiest way to reproduce it, but unlikely to be the only one. [This VSCode documentation](https://code.visualstudio.com/api/extension-guides/custom-editors#custom-text-editor) describes how we should be behaving.

Open question: Are we doing (or not doing) anything that clashes with the way we are supposed to do it?

PR #98, opened 2024-11-01
Temporary fix for undo problem.

DikieDick:
### Description
Designed as a fix for #97: We add a custom `HistoryChange` edit to our edit queue whenever the user triggers either `waterproof.undo` or `waterproof.redo` using the keybindings `ctrl-z` (`cmd-z`) and `ctrl-y` (`cmd-y`) respectively. This `HistoryChange` edit is added to the queue and upon execution, triggers the regular `undo` and `redo` command.

### Changes
- Added keybindings for `waterproof.undo` and `waterproof.redo` that are only active when the waterproof editor is in focus.
- Changed `handleChangeFromEditor(change: DocChange)` to also accept `change: HistoryChange` and added an implementation for these two edits.

### Testing this PR
Use the editor and change library imports. Don't let the editor 'stabilize' and press `ctrl-z`. The editor state should remain synchronized.

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,16 +148,16 @@
         "when": "editorTextFocus && editorLangId == coq || editorTextFocus && editorLangId == coqmarkdown"
       },
       {
-        "command": "waterproof.undo", 
+        "command": "waterproof.undo",
         "key": "ctrl+z",
         "mac": "cmd+z",
-        "when": "editorFocus && activeCustomEditorId == coqEditor.coqEditor"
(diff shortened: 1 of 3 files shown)


jim-portegies (changes requested):
Thanks so much! I've tested it and it fixes the problems that I know of. There is one issue, that when opening a second Waterproof file, it says: "Failed to create editor view" "Error: the command 'undo' already exists".

jim-portegies (approved):
Thanks so much! I think this looks good. I did quite a bit of testing and could not break it anymore.

PR #106, opened 2024-11-05
Move to specified versions, revert @codemirror/view to 6.27.0

pimotte:
### Description
Fixes [#105](https://github.com/impermeable/waterproof-vscode/issues/105)

### Changes
Moved to specifically pinned versions in package.json, reverted @codemirror/view to 6.27.0.

In general, in npm projects, I like to use the following convention: For libraries, use wide dependency ranges, since projects that use it might have a satisfactory version in the tree already. For end-user applications, specify the dependencies relatively tightly. This makes updating slightly more effort, but also makes it easier to track down and fix stuff like this.

The changelog mentions a switch to the EditContext API in version 6.28.0 for Chrome, I think we are in a Chrome-like environment, so this might be why it breaks for us.

### Testing this PR
Run `npm ci`
Run the editor. Copy and paste stuff. Cut and paste stuff.
Possibly test for regressions, but this is just changing this dependency and a minor difference in react versions due to the pinning, so I think it should be fine in general.


DikieDick (approved):
Everything seems to be working as intended.

jim-portegies:
At the moment, the menu version of copy, paste, cut still do not seem to work. This may not be critical.

pimotte:
> At the moment, the menu version of copy, paste, cut still do not seem to work. This may not be critical.

Did you run `npm ci` before testing? It works fine on my end. (Yes, I did edit the testing instructions after your comment:P)

jim-portegies:
Yup, I did run `npm ci`, mystery why it doesn't work.

pimotte:
Then I'm going to boldly assume that may be an orthogonal issue. I've created [this card](https://github.com/orgs/impermeable/projects/5/views/6?pane=issue&itemId=86154992) for it.

PR #107, opened 2024-11-05
Forward undo/redo events to ProseMirror, ProseMirror maintains undo/redo stack.

DikieDick:
### Description
Let ProseMirror maintain its history stack and forward undo/redo events to the editor instead of relying on the vscode undo and redo implementations.

This fixes the flickering of the screen when doing undos/redos, does not interrupt the edit workflow on `ctrl-z` and generally makes the editor feel more polished.

Builds on top of #98.

### Changes
Upon undo/redo forward a message to the editor, informing that it should perform an undo or redo operation.

Requires the `prosemirror-history` package.

### Testing this PR
Also see #98. Try using the editor as one normally would, make some typos and make heavy use of the undo and redo actions. It should work with the relevant shortcuts (`ctrl-z`/`cmd-z` and `ctrl-y`/`cmd-y`) as well as with the menubar "edit > undo" or "edit > redo" actions. 

Furthermore, try to think of edge cases, for example open more than one file, switch between them and make edits in both that you can then undo.


jim-portegies:
When editing text, the new undo seems much more natural. However, it seems that creating a new code cell cannot be undone with the new undo, while it could be with the old undo.

pimotte:
I tested this locally (but I didn't have the original issue), works well for me:)

I also took the liberty of backmerging main:)

DikieDick:
I can look into the undoing of code cells a bit more, I suspect that ProseMirror creates edit that we do not support yet.

DikieDick:
Thanks for merging main @pimotte, but the `HistoryChange` input (and corresponding logic) to `applyChangeFromEditor` needs to be removed, the rest looks fine.

Shall I do this?

jim-portegies (approved):
Thanks! The user experience is indeed much better.

In a later stage, we may want to support undo-ing the creation of cells.

PR #108, opened 2024-11-05
Custom placeholder for coq cells placed within an input area.

DikieDick:
### Description
The custom placeholder says `(* Type your proof here *)` and can be styled separately from the regular placeholder.

_Note_: There is no reason for it to look like a comment anymore using this approach. Different text like `Start typing your proof here...` would work just as well and not cause a syntax error in the document.

As this placeholder is not part of the text, the user can click anywhere within the cell and start editing, without having to worry about removing the comment.

### Changes
In `nodeview.ts` (where we create the placeholder) we determine whether the cell is part of an input area. If so, we return a div with text `(* Type your proof here *)` and class `empty-proof-placeholder`, if not, we return the old placeholder text wrapped in a div.

The styling can be customized in `editor/src/kroqed-editor/styles/input-area.css`.

As a demonstration I have removed the `(* Type your proof here *)` comments from the tutorial.

### Testing this PR
Start Waterproof and open a fresh tutorial instance. There should be no visual difference with how the tutorial looked before this PR.

When you now click on any part of the placeholder, you can start editing right away without having to remove the comment.


pimotte (approved):
Looks good to me! I support keeping the placeholder as-is. This way it immediately serves as a pointer how students could include comments if they wanted to. 

Note that this is also backwards compatible: Files that contain an actual comment as placeholder cause no issues, so I think this is fine to merge.

jim-portegies (approved):
Thanks a lot! I believe this improves the user experience!

I have just updated saving to exercise sheet accordingly.

jim-portegies:
Thanks also to @pimotte for the idea.

PR #109, opened 2024-11-05
tab functionality added to indent, if multiple lines selected only fi…

Alisakondratyev:
…rst line indented and accept autocomplete

### Description
updated tab behavior. Tab can be used to indent, if multiple line selected then only the first line will indent (temporary fix). Tab also can be used to accept an autocomplete suggestion.

### Changes
A custom tab function, customTab was created in embedded-codemirror-keymaps that accepts a completion if there is one to select, otherwise indents the line.  The codemirror commands indentMore and changeBySelectedLine are modified to only indent the first line of the selection. In EditableView and additional customTab function is added that uses the codemirror command indentMore, as the keymap I defined for tab in embedded-codemirror-keymaps did not work in markdown cells.


jim-portegies (approved):
Thanks a lot!

The behavior of the autocomplete with tab is nice. I also think the solution for tabbing is reasonable for now.

At the moment, this doesn't seem to respect the "Editor: Accept Suggestion On Enter" setting yet, but let's fix that in a separate PR.

PR #118, opened 2024-11-18
Temporary fix for <5 in mapping input content

DikieDick:
This temporary fix creates an edit that adds a space between the less than sign and the character following it, except if it is a known tag in our model ('hint', 'input-area', etc).

### DIFF
--- a/__tests__/editor-content-construction.test.ts
+++ b/__tests__/editor-content-construction.test.ts
@@ -110,6 +110,30 @@ test("TEMP: Leave <hr> in markdown untouched #2", () => {
     }]);
 });
 
+test("TEMP: Works with multiple < signs", () => {
+    const pm = new PostManager();
+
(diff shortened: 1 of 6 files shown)


jim-portegies (commented):
Thanks! But unfortunately I still have issues. If I fill in Exercise 7 of the tutorial with:
```
Assume that ( f(3) < 5 ).
By (f_increasing) we conclude that (& f(-1) ≤ f(3) < 5).
```
and in exercise 8 I have
```
Assume that (f(5) < 4 ).
```
and then I remove the spaces after `<` in our editor with a backspace, save and reopen the file, the second line is changed to 
```
Assume that (f(5)  <4 ).
```
I'm not sure if we should merge this now or wait until we find a fix for this.

jim-portegies:
The backspace is not essential, if I type it out, I get the same problem unfortunately.

DikieDick:
Please also test and review #119 which provides a neater solution to the same problem.

jim-portegies:
@DikieDick do we still want to merge a version of this that is actually not triggered? Or leave it for now?

jim-portegies:
Closing this, since #119 is a better solution.

PR #122, opened 2024-11-21
Proof of concept list of tactics

DikieDick:
### Description
I don't remember when but somebody asked me if there was a list of available Waterproof tactics. We do have the tactics panel, but that only works in the extension and (in my opinion) is a bit cramped/ doesnt read nicely. 

This PR adds a script that we could run in the CI to generate a MD file / PDF file of the available tactics. 

### Changes
- Add scripts folder and add this folder to the `.vscodeignore` file.
- Adds script to generate a list of all available Waterproof tactics in Markdown format. See [here](https://github.com/impermeable/waterproof-vscode/blob/generate-pdf/scripts/tactics.md).

### Note
Using `pandoc` we can turn this into a webpage (`pandoc -s tactics.md -o tactics.html`) or a pdf document.


pimotte (approved):
I think this is a great idea, I think it's just missing some discoverability. Maybe link it from the readme, taking care it also works from in the extension? And maybe put the output in a `docs` folder?

I'm good with merging it as-is if you're just doing some cleanup, if you don't end up doing the above, can you create an issue for that?

Also, this will likely get merged before #184 and we'll need to rerun the script there.

DikieDick:
I will hapilly incorporate the changes

pimotte:
Looks good to me:)

PR #125, opened 2024-12-03
Updated goals panel

XyntaxCS:
### Description
Removes lengthy description of remaining goals and subproofs. Also removes autoscrolling to the bottom of the goals panel.

### Testing this PR
Check some files that have lengthy proofs with many steps that would normally require subproofs. The example I used is exercise sheet 5. This contains numerous subproofs. Please use another file to avoid it just being this case that works.


views/goals/Goals.tsx
```diff
@@ -136,6 +176,8 @@ export function Goals({ goals, pos, textDoc }: GoalsParams) {
       No goals at this point.
     </Box>
   }
+  console.log("test")
+  console.log(goals)
```
jim-portegies, line 180:
```suggestion
```

views/goals/Goals.tsx
```diff
@@ -107,23 +135,35 @@ function StackGoals({ idx, stack, pos, textDoc }: StackSummaryP) {
...
-        {/* recursively calls itself to get the different level of goals */}
-        <StackGoals idx={idx + 1} stack={stack} pos={pos} textDoc={textDoc} />
+  console.log(idx)
+  if (idx == 0) {
+    return (
+      <div>
+        {/* uses GoalsList to show the goals as a list within a stack */}
+        <GoalsRemaining
+          goals={goals}
+          header={`Remaining subproofs/steps (some statements/cases remain to be shown)`}
```
jim-portegies, line 145:
```suggestion
          header={`Afterwards, we need to complete other subproofs`}
```

views/goals/Goals.tsx
```diff
@@ -107,23 +135,35 @@ function StackGoals({ idx, stack, pos, textDoc }: StackSummaryP) {
...
-  );
-}
+    );
+  } else {
+    return (
+      <div>
+        {/* uses GoalsList to show the goals as a list within a stack */}
+        <GoalsRemaining
+          goals={goals}
+          header={`Remaining subproofs/steps (some statements/cases remain to be shown)`}
```
jim-portegies, line 158:
```suggestion
          header={`Afterwards, we need to complete other subproofs`}
```

jim-portegies (changes requested):
Already a big improvement! Thanks!

It seems that sometimes if there are multiple goals, and they are not focused yet, then all goals are printed. Would it be better to only show the first goal?

jim-portegies (approved):
Thanks a lot! I think this makes things a lot clearer. I brushed up the notation a bit: fixed two typos and removed a remark about foldable goals because I think we currently don't fold.

PR #127, opened 2024-12-10
New diag buttons

raulTUe:
### Description
Diagnostic Buttons Replace and Insert have been added, to allow for users to use diagnostics for their proof at the click of a button. Now focus on a codeblock is no longer needed, and correct indentation has been applied. 
### Changes
A new updateDiagnostics method has been added, which ensures that the buttons are updated based on all the diagnostics in the line. If there is an error, then a replace button appears for all diagnostics, and if there is no error then the insert button appears for all diagnostics. Moreover, the copy button now has some responsiveness, and tells the user when the text has been copied to their clipboard.
### Testing this PR
You can test it by trying different syntax errors or by inputting relevant tactics in the editor (as seen in the tutorial section 17). You can also place your cursor on a different codeblock and click the diag buttons to test the focus.


jim-portegies (changes requested):
The feedback on copying is now much better, but maybe we should stay closer to the color scheme that we are using. Can we make use of one of the theme colors?

Beyond that, the indentation still doesn't work properly. If I write down
```
  Expand the definition of supremum.
```
and click "Replace", then it gets four leading spaces instead of two. If it has four leading spaces to begin with, it receives eight leading spaces after replacing.

jim-portegies:
In addition, I think we should not change the `vscode/settings.json` file

pimotte (changes requested):
Two small nitpicks, tested and behaviour looks good.

DikieDick (approved):
Approved by me. I wanted to comment that the styling could be centralized to the `styles.css` file imported in `editor.ts` but @pimotte has already fixed this.

PR #128, opened 2024-12-18
Autocomplete tactic fix

Alisakondratyev:
### Description
This provides a rudimentary fix for the issue of accepting one tactic suggestion, partially deleting the suggestion, and then accepting another suggestion. Previously this would cause the new tactic to be inserted part of the way through the line.

### Changes
I have now set the tactic suggestions to always start from the beginning of the line explicitly.


jim-portegies (commented):
Thanks! I do think this solves some issues. When testing, I noticed that the autocomplete now also doesn't work when there is indentation, or when a line starts with bullet points.

jim-portegies:
@Alisakondratyev would it be possible to resolve the merge conflict?

pimotte (approved):
Fixed some braces (Might have gone wrong during the merge) and after it worked nicely:)

PR #134, opened 2025-01-13
Update version to 0.9.18

jim-portegies:
### Description
Just a change a version number preparing a new release.

### Changes
Version number changed from 0.9.17 to 0.9.18

### Testing this PR
Double check if 0.9.17 is no longer present in the code, and is changed to 0.9.18 where it was present

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "waterproof",
-  "version": "0.9.17",
+  "version": "0.9.18",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "waterproof",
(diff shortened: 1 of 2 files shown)


pimotte (approved):

pimotte:
Maybe it makes sense to try and merge some of the outstanding PRs before releasing? I think #124 #127 #128 and possibly also #125 would be really nice to have.

jim-portegies:
@pimotte  In fact I'd like to release this version because it is close to the version that's tested in class + a bug fix. I would like to include the new features in a new release.

pimotte:
That makes sense too, works for me:)

jim-portegies:
Ok, thanks!

PR #144, opened 2025-01-28
Checking input areas

### LINKED ISSUE #115: Input areas are not checked or remain blue despite being checked.

raulTUe:
### Description
Fixes #115
### Changes
Asynchronously wait for the client to have initialized before focus event.
### Testing this PR
Green/red bars in sidelines are shown as soon as coq-lsp finishes checking, instead of after first edit.


pimotte (approved):
I can't really reproduce the original issue consistently, but on main there's also a little bit of a delay when starting up and as I switch editors I notice no noticable delay. The promise-construction seems the correct one to me and the logging should be enough to debug this in the wild if it gets stuck here, so I'd actually be good with merging this as-is:)

jim-portegies:
I relatively consistently get the following behavior:
* Before I close vscode, I make sure `limits_and_continuity_13_11_2.mv` is still open
* I reboot vscode and then open in addition `construct_index_sequence.mv`
* `construct_index_sequence.mv` gets checked twice, after which all input areas are colored blue. If I then check `limits_and_continuity_13_11_2.mv`, the visuals seem to suggest that the checking is stuck at some line, but by removing and adding a period, the file is immediately completely checked.

pimotte:
My proposal: @raulTUe will make sure that this PR fixes the original issue, and we split off the issue @jim-portegies raised to a separate one.

jim-portegies (approved):
At the moment, I cannot really reproduce the original issue on main. But it still seems to work on the new branch. So it's probably good to merge it.

@raulTUe should we remove the log messages?

PR #158, opened 2025-03-17
Make extension web-compatible

pimotte:
### Description
Ensures we can build and use the web-compatible

### Changes
- Setup web worker for LSP
- Remove usage of node.js API and replace with vscode API or browser/node cross compatible packages

### Testing this PR

- Running the webextension requires copying `coq_lsp_worker.bc.js` to the `./out` folder. (From coq-lsp [build artifact](https://github.com/ejgallego/coq-lsp/actions/runs/13566988935)
- Running the webextension can be done using the "Run Web Extension in VSCode" launch target, or by running `npx @vscode/test-web --extensionDevelopmentPath=.` in the root of your project. The latter is a lot more performant.
- In testing this, mostly everything needs a spot-check in the web version. For the normal version we need to check everything that runs a command, in particular I cannot test the autoinstaller.


pimotte:
To fully enable the webextension, we need to either:
- Download the `coq_lsp_worker.bc.js` file in the build process (annoying, because it's not a public download, requires github auth of some sort)
- Vendor `coq_lsp_worker.bc.js` in our git (really don't want to do this, it's 188MB).
- Use the extension API of coq-lsp to obtain our webworker (long term plan)

Because there are already a decent number of changes in this branch, I'd like to merge this and follow up later with the PR that does the third option above, to prevent stuff from diverging too much.

DikieDick (commented):
I should be testing with Coq 8.20, right?

Side note: Opening a new waterproof file using the "Create new Waterproof document" command, throws an error (edit: as does open tutorial)
.

src/helpers/logger.ts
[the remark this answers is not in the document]
pimotte: Correct, I just introduced the split to be able to do this later.

src/extension.ts
```diff
@@ -298,14 +296,18 @@ export class Waterproof implements Disposable {
      */
     private async autoInstall(command: string): Promise<boolean> {
         return new Promise((resolve, _reject) => {
-            exec(command, (err, _stdout, _stderr) => {
-                if (err) {
-                    // Simple fixed scripts are run, the user is able to stop these but they are not considered errors
-                    // as the user has freedom to choose the steps and can rerun the command.
+            const myTerm = window.createTerminal(`AutoInstall Waterproof`)
+            window.onDidChangeTerminalShellIntegration(async ({ terminal, shellIntegration}) => {
```
DikieDick, line 300: This event never fires for me on Windows. So although we open a terminal (which doesn't show by default, so we should consider calling `myTerm.show()` in my opinion), nothing happens after that.

DikieDick (changes requested):

src/extension.ts
pimotte: This was a pretty nasty race condition, so happy you caught it:D It should be fixed now

DikieDick: For me on Windows nothing happens when we hit `await this.stopClient` in the `autoInstall` register function. I don't have the time to debug it much further now, but on Linux it seems fine. 

I tested this by printing before and after the await call and only the before print ever showed up on Windows. Furthermore, the WaterproofLogger does not behave correctly on Windows I think (maybe this is related?)

pimotte: If you can spend a little time on this later, that would be great, if it becomes too much, let me know, and I can see if I can setup a windows environment for this stuff:)

raulTUe:
@DikieDick I'm going to look into testing the most recent PR, no need for you to.

pimotte:
@raulTUe FYI, I have found out that `onDidChangeTerminalShellIntegration` just doesn't fire on windows for PowerShell 5. According to #226655 it "should work", but I'm not really finding any information on how this could be broken. It does work for PowerShell 7, but I don't think this works, so I'll try to revert this to something that uses dynamic imports to circumvent the webextension issues.

pimotte:
@DikieDick Could you re-review this? :)

DikieDick:
First comment before the review, it seems that opening the tutorial and a new document still does not work.

src/extension.ts
```diff
@@ -330,27 +331,25 @@ export class Waterproof implements Disposable {
     }
 
     private async waterproofTutorialCommand(): Promise<void> {
-        const defaultUri = workspace.workspaceFolders ?
-            Uri.file(joinPath(workspace.workspaceFolders[0].uri.fsPath, "waterproof_tutorial.mv")) :
-            Uri.file(joinPath(homedir(), "waterproof_tutorial.mv"));
+        const defaultUri = Utils.joinPath(workspace.workspaceFolders![0].uri, "waterproof_tutorial.mv");
```
DikieDick, line 334:
```suggestion
        const hasWorkspaceOpen = workspace.workspaceFolders !== undefined && workspace.workspaceFolders.length != 0;    
        const defaultUri = Utils.joinPath(workspace.workspaceFolders![0].uri, "waterproof_tutorial.mv");
```

Using something like this.

pimotte: My bad, adopted this, supplying a default name without any specific path if there is no open workspace

src/helpers/file.ts
```diff
@@ -0,0 +1,21 @@
...
+        let base = new String(filePath.path).substring(filePath.path.lastIndexOf('/') + 1); 
+        if(base.lastIndexOf(".") != -1)       
+            base = base.substring(0, base.lastIndexOf("."));
+        return base;
+    }
+
+    public static join(...paths: string[]): string {
+        const sep = process.platform === "win32" ? "\\" : "/";
+        return paths.join(sep);
+    }
```
DikieDick, line 20:
```suggestion
        // We filter out empty path strings, maybe we could instead make this function
        // assume that arguments are non-empty and let the caller handle these?
        paths = paths.filter(v => v != "");
        const sep = process.platform === "win32" ? "\\" : "/";
        return paths.join(sep);
    }
```

DikieDick:
The web version of the extension seems to work as described. Is it correct that we currently do not get any goals in the goals panel when running the webextension?

DikieDick:
I'm fine with leaving the web version as is.

pimotte:
> I'm fine with leaving the web version as is.

Interpreting this as "approved" and merging this:)

PR #162, opened 2025-03-25
Refactor editor

DikieDick:
### Description
Collection of changes to abstract away from Waterproof and Coq specific usage.

As an example, we abstract away the usage of message within the editor itself. In a fully browser use case one would just call `WaterproofEditor.x` to tell the editor to do something instead of sending a message. 

This PR is part of a bigger refactor that I want to do *before* splitting off the editor, but I'm currently waiting on the mapping to be finalized by @XyntaxCS.

### Changes
- Rename the editor class from  `Editor` to `WaterproofEditor`.
- Move *all* messages to a file responsible for communication with the extension.
- Remove references to `Coq`.
- Remove DOM specifics from the ProseMirror schema.

### Testing this PR
Test the extension by going through different files.


editor/src/kroqed-editor/schema/schema.ts
```diff
@@ -145,38 +127,29 @@ export const TheSchema: Schema = new Schema({
...
-			}]
 		},
 		//#endregion
 	},
-	marks: {
-		em: {
-		  parseDOM: [{tag: "i"}, {tag: "em"}],
-		  toDOM() { return ["em"] }
-		},
+	// marks: {
```
pimotte, line 133: Do we want to keep this?

editor/src/kroqed-editor/editor.ts
```diff
@@ -257,15 +252,15 @@ export class Editor {
 			menuPlugin(this._schema, this._filef, this._userOS),
 			keymap({
 				"Mod-h": () => {
-					this.post({type: MessageType.command, body: { command: "Help.", time: (new Date()).getTime() }});
+					this.executeCommand("Help.");
```
pimotte, line 255: This code is supposed to be generic, right? This seems to still be tied to the waterproof syntax.

pimotte (commented):
LGTM, two minor comments. Spot checked the editor. I'm pretty confident I have hit all API methods except `applyStepError`, which I think is luckily hard to trigger now:P

editor/src/kroqed-editor/schema/schema.ts
DikieDick: They are not in use at the moment. The marks are remainders of the [`prosemirror-markdown`](https://github.com/ProseMirror/prosemirror-markdown/blob/master/src/schema.ts#L118) schema that we used before switching to custom markdown rendering.

editor/src/kroqed-editor/editor.ts
DikieDick: This will indeed be part of the later PR

jim-portegies (approved):
Thanks for this work!

I have worked on several files, also tried to insert code cells etc, and things seemed to be okay!

PR #164, opened 2025-04-14
Simple fix for side panel

Alisakondratyev:
### Description
This has a really simple fix for the behavior of the goals panel on starting the extension. Previously the logbook, and sometimes debug buttons would be greyed out. Now only the goals side panel should be opened and the button greyed out. 

### Changes
I removed two lines that were causing the issue.

### Testing this PR
When I run the extension now, the behavior is as expected (only goals panel opened).


pimotte (approved):
Can confirm the debug button with respect to main. (Logbook weirdly wasn't grey on main).

With both main and the current change the goals button is enabled even though the panel is open, but I don't think this PR is supposed to resolve this.

@Alisakondratyev If the above is correct, feel free to press the merge button:)

PR #166, opened 2025-04-22
Fixing button transparancy for side panels

Alisakondratyev:
### Description
This provides a fix for the issue of button transparency in the side panels. Now, when a user reopens the side panel or opens a new document, the opened panels are saved, and the buttons are greyed out. Previously, when a user would reopen the side panel, all the buttons would be reset to their original state, even if panels were opened.

### Changes
I added an event listener for changing the panel visibility (onDidChangeVisibility), as well as a function that keeps track of the buttons that are greyed out (greyedOutButtons). When the side panel is reopened (when onDidChangeVisibility), the side panels that were opened (those in greyedOutButtons) are greyed out.

### Testing this PR
I have tried testing these changes by opening and closing the side panel after clicking open different side panels and by opening new exercise sheets.

### DIFF
--- a/src/webviews/sidePanel.ts
+++ b/src/webviews/sidePanel.ts
@@ -61,7 +61,7 @@ export class SidePanelProvider implements vscode.WebviewViewProvider {
 
     public resolveWebviewView(
         webviewView: vscode.WebviewView,
-        context: vscode.WebviewViewResolveContext,
+        _context: vscode.WebviewViewResolveContext,
         _token: vscode.CancellationToken,
     ) {
         this._view = webviewView;
@@ -101,7 +101,7 @@ export class SidePanelProvider implements vscode.WebviewViewProvider {
     }
 
     // Now we create the actual web page
-    private _getHtmlForWebview(webview: vscode.Webview) {
+    private _getHtmlForWebview(_webview: vscode.Webview) {
         const nonce = getNonce();
 
         // html code for the webview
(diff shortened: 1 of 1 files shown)


pimotte (changes requested):
Work well, with the exception of the bug I mentioned. A little refactoring will make this more maintainable, that's what my suggestions aim towards:)

pimotte:
@Alisakondratyev Because of the line `panel?.activatePanel()` I had a sudden realisation: What if we don't grey out the buttons at all, but instead have a press of the button reveal the panel. We already have a `revealPanel` function.

The common usage pattern is that people close or hide their goals panel and then go here to get it back. When it's greyed out they then know (or need to be explained) that this means the panel is already open.

You've spent a lot of time around this feature, do you think this makes sense (also keeping in mind the other problems you've been fighting in this area?)

Some advantages I can think of:
- Improves the user flow (instead of having to go look for the panel, they just click the button and get it back)
- No need to keep track of the state of what is or is not open.

Possible disadvantages:
- Technical limitations I'm not aware of (Maybe we want to specificy a column in the `reveal()` call in `revealPanel()`
- Prior reasons why we do it the current way (@jim-portegies Thoughts?)

jim-portegies:
I think that makes sense!

pimotte:
My guess is that we don't need to keep track of that: you can query for the panel, if you find it, reveal it, else open it

PR #169, opened 2025-04-24
Fix bug in handling of help commands

DikieDick:
### Description
After refactoring the messages, their was an inconsistency between the messages used for triggering the `Help.` command.

This has now been fixed by fixing the message that is sent to the `WebviewManager`

### Changes
- `src/webviewManager.ts`: The command that we want to execute is stored in the `command` field of the body. The body itself is an object, this has been updated.
- `views/help/help.tsx`: Split `msg: Message` to allow for typechecking of the message. 

### Future Work
We should either remove the `WebviewManager` messages or make sure that they are also properly typed to catch these kinds of bugs.

### Testing this PR
Go through (parts of) the tutorial as usual.
Ask Waterproof for help using the various options available:
- The `help` button in the `Help` panel.
- Clicking the right-mouse-button followed by `? Help.`
- Using the key combination `ctrl-h` (`cmd-h` on Mac) in the editor.

### DIFF
--- a/src/webviewManager.ts
+++ b/src/webviewManager.ts
@@ -271,7 +271,8 @@ export class WebviewManager extends EventEmitter {
                 this.postMessage(id, { type: MessageType.insert, body: msg.body });
                 break; }
             case MessageType.command:
-                this.emit(WebviewManagerEvents.command, this._toolWebviews.get(id), msg.body);
+                // FIXME: The `WebviewManagerEvents` are **not** typed.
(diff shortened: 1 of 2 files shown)


jim-portegies (approved):
Thanks for the fix! I tried around in the tutorial and it seems to work fine.

Is there a test we could add for this?

On right-click, the help message does not show up immediately. I thought we had fixed that at some point but I cannot recreate it in earlier commits either, so it is okay, I might make a new issue.

PR #171, opened 2025-04-30
Trimming whitespace fix

raulTUe:
### Description
This is a pull request to partially solve the trimming whitespace issue. We now reload the editor only if the file has changed on save, and always give a warning to the user when the trim trailing whitespace setting is on.
### Changes
The editor effectively re-initializes if the underlying file has changed when the file was saved, which only really occurs when the trim trailing whitespace setting is on. This makes it clunky if auto save is on, but the warning should prevent this.
### Testing this PR
When initializing the extension, the warning should appear on the bottom left of the screen. Moreover, if you turn on trim trailing whitespace setting, and then add a couple of spaces at the end of one of the input areas, and press ctrl+s, then the editor will flash and reinitialize.


editor/src/kroqed-editor/editor.ts
```diff
@@ -150,6 +150,47 @@ export class WaterproofEditor {
 		this._editorConfig.api.editorReady();
 	}
 
+	updateDocument(content: string, version: number) {
```
pimotte, line 153: I'm tempted to say "refreshDocument" or "reloadDocument" better reflects what this actually does.

editor/src/kroqed-editor/editor.ts
```diff
@@ -150,6 +150,47 @@ export class WaterproofEditor {
...
+		}
+		if (resultingDocument !== content) version = version + 1;
+
+		const parsedContent = this._translator.toProsemirror(resultingDocument);
+		const proseDoc = createProseMirrorDocument(resultingDocument, this._filef);
+
+		this._mapping = new TextDocMapping(this._filef, parsedContent, version);
+		this.createProseMirrorEditor(proseDoc);
+
+		this.sendLineNumbers();
```
pimotte, line 189: Refreshing unfocuses the code cell and removes the cursor. This happens when you stop typing after you type a space.

I can't remember if we decided this was acceptable. It is acceptable to me, because the students shouldn't need trim spaces, but I just want to check with @jim-portegies if you agree.

pimotte (commented):
Works well, some minor comments:)

editor/src/kroqed-editor/editor.ts
DikieDick: +1

editor/src/kroqed-editor/editor.ts
jim-portegies: If there's an easy fix for the cursor problem, it may be worth including it

editor/src/kroqed-editor/editor.ts
DikieDick: I also think it would be nice if you can add a little documentation explaining what this function does.

shared/Messages.ts
```diff
@@ -34,6 +34,7 @@ export type Message =
     | MessageBase<MessageType.editorReady>
     | MessageBase<MessageType.errorGoals, unknown>
     | MessageBase<MessageType.init, { value: string, format: FileFormat, version: number }>
+    | MessageBase<MessageType.updateDocument, {value: string, version: number}>
```
DikieDick, line 37: Can you also update the documentation of the messages in `developer-resources/messages.md`?

DikieDick (changes requested):

DikieDick (commented):
The check for the `trimTrailingWhitespace` setting works well.

I noticed that there is some problem regarding existing diagnostics in the document, however:
1. Open the tutorial and navigate to the first input area.
2. Enter `Test.   ` (mind the trailing spaces).
3. Hit `Ctrl-S` to save and hence refresh the document.
4. Replace `Test.` by `We conclude that (3 = 3).`
5. The old syntax error regarding `Test.` reappears.

raulTUe:
With this implementation there still seems to be a bug when saving files with the trim whitespace setting on:
1. Open the tutorial and go to the first input area.
2. Type Test. (mind the trailing spaces).
3. Hit Ctrl-S to save and hence refresh the document.
4. Click again in the same input area, where Test. lies without the trailing whitespace.
5. Hit backspace once, and wait 5 seconds.
6. Try hitting backspace or typing any other letter.
7. Any input in this input area doesn't have any effect.

pimotte:
Is superseded by #345

PR #181, opened 2025-07-02
Add option to disable launch checks

### LINKED ISSUE #180: Feature: Option to turn off version checks
### Is your request related to a problem? If so, please explain.
When people want to use the waterproof interface with plain rocq, they won't need coq-waterproof.
In addition, sometimes the version check misbehaves and says the install is not ok, when it actually is.

### I would like.
To add an option that turns off all the version/livenesschecks of the LSP. In the option we should explain that it is then the users responsibility to ensure that a coq-lsp with necessary dependencies installed is available.

Acceptance criteria:
- A new option "Skip version checks" is added, default is False
- When this option is true, we do not check for: presence of coq-lsp and versions of rocq related projects. Instead we just attempt to start coq-lsp.

DikieDick:
### Description
Adds a setting that disables the launch checks. 

Launch checks normally check for:
- `coq-lsp` installation
- `coq-waterproof` installation and version

Disabling comes at the risk that we try to initialize the client when this is not possible, hence the warning message in the setting.

Closes #180.

### Changes
- Updated `package.json` and `WaterproofConfigHelper` to include new setting. 
- Don't run launch check when the option has been set.

### Testing this PR
1. Run the extension like you normally would. Everything should function as expected.
2. Use `opam switch` to switch to a switch without `coq-waterproof`, without the `skipLaunchChecks` option set this should result in an error message. After turning the option on, the client should start normally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -321,6 +321,11 @@
             "type": "boolean",
             "default": false,
             "description": "Show line numbers in the Waterproof editor"
+          },
+          "waterproof.skipLaunchChecks": {
+            "type": "boolean",
+            "default": false,
+            "description": "Disables the checks that happen when launching the extension. During those checks Waterproof checks whether Coq, coq-lsp and coq-waterproof are installed. Use at your own risk!"
           }
         }
       },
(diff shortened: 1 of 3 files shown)


pimotte (approved):
Looks good to me:D

PR #188, opened 2025-07-09
Show hypotheses feature toggle

### LINKED ISSUE #179: Feature: Show hypotheses option
Currently, we purposefully don't show the hypotheses.

To make the waterproof editor useable for education using plain Rocq, it is desirable to be able to show the goals. (Feature request from @gmalecha)

I would argue that we want to allow this, even at the risk of students discovering this and turning it on. @jim-portegies just to check, I think you agreed with me on this, but if not, please let me know.

Work in progress branch [here](https://github.com/impermeable/waterproof-vscode/tree/feature/show-hypotheses)

Acceptance criteria:

- Option "Show hypotheses" is added to waterproof settings with options "None", "Limited" and "All"
- If option is turned "None", current behaviour is retained
- If option is "Limited", all hypotheses that don't start with `_` are shown in the goals panel.
- If option is "All", all hypotheses are shown

### LINKED ISSUE #186: [Bug] Debug is broken
### What happened?
The debug view doesn't show anything

1. Open tutorial
2. Open logbook
3. See logbook stuck on "Loading"

Show the goal and hypotheses

Ran into this while refactoring renderGoals for #179

pimotte:
### Description
Adds feature toggle to show all, limited or none (default) hypotheses.
Fixes the debug panel.

### Changes
- Splits the renderGoals message type into renderGoals and renderGoalsList to allow typing them.
- Resolves #179 
- Fixes #186 
- Manually reverts #164 (This broke debug and possibly logbook functionality).

### Testing this PR
Test the different settings and check if the hypotheses are shown to the appropriate level.

### Open points

- What is the logbook supposed to show? It should be rendering the "messages", but these are empty whatever goal I inspect.
- This breaks the button transparency again1

### DIFF
--- a/views/goals/Messages.tsx
+++ b/views/goals/Messages.tsx
@@ -6,6 +6,7 @@ import {Box} from "./Box";
 import "../styles/messages.css";
 import { PropsWithChildren } from "react";
 
+
 //type that makes a GoalAnswer<PpString> also takes its childrens components with
 export type MessagesInfo = PropsWithChildren<
   {
@@ -16,6 +17,7 @@ export type MessagesInfo = PropsWithChildren<
 //component that takes in the MessagesInfo and displays the list of messages
 export function Messages({answer} : MessagesInfo) {
   const count = answer.messages.length;
+  console.log("Messages: count is ", count);
   //check if there are any messages that need to be shown
   if (count != 0) {
   //the Box component is used to display the messages along with the location
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -143,3 +143,9 @@ export interface DocumentPerfParams {
     summary: string;
     timings: SentencePerfParams[];
 }
+
+export enum HypVisibility {
+  All = "all",
+  Limited = "limited",
+  None = "none"
+}
--- a/package.json
+++ b/package.json
@@ -393,6 +393,16 @@
               "Use jsCoq's Pp rich layout printer",
               "Coq Layout Engine (experimental)"
             ]
+          },
+          "waterproof.hyp_visibilty" : {
+            "type": "string",
+            "default": "none",
+            "enum": [
+              "all",
+              "limited",
+              "none"
+            ],
+            "description": "Visibility of hypotheses in the goals panel.\n- `all`: show all hypotheses\n- `limited`: show only hypotheses not starting with _\n- `none`: do not show any hypotheses\nFor didiactic purposes, we recommend `none` for mathematicians learning proofs and `limited` or `all` for students using the Rocq language."
           }
         }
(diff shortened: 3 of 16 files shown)


pimotte:
After conferring with @jim-portegies, we have come to the conclusion we can drop the logbook functionality. The error boxes replace the main use case: Being able to easily see all errors in a document.

views/goals/Messages.tsx
```diff
@@ -16,6 +17,7 @@ export type MessagesInfo = PropsWithChildren<
 //component that takes in the MessagesInfo and displays the list of messages
 export function Messages({answer} : MessagesInfo) {
   const count = answer.messages.length;
+  console.log("Messages: count is ", count);
```
DikieDick, line 20: Do we need this?

DikieDick (approved):
I think  we should add a separator in between the hypotheses and the goal, other than that looks good to me and works very nicely.

pimotte:
I ended up moving things around a little bit, to keep the goal with the "We need to show" and added a "Context:" header for the hypotheses section. I'm not really sure about the verbiage here. I think "Context", "Hypotheses" and "Environment" are all valid options. @jim-portegies Any preference? (Also, screenshot below for a final visual check, this uses the display all hypotheses option.)

image

views/goals/Messages.tsx
pimotte: Fixed

jim-portegies (commented):
Thanks!

A few remarks:

In the settings menu, the rendering of this setting can be a bit nicer. I recommend to describe the setting there in a way different from "Hyp_visibility", but just with a few words, such as "Visibility of hypotheses". Moreover, the list does not get rendered very nicely, as the backticks are just visible. Finally, but this is maybe not so much an issue here, but more for the Waterproof theme, in the Waterproof theme the drop down menu cannot be read.

I am thinking maybe we can use "We know:" instead of "Context".

pimotte:
> Thanks!
> 
> It looks very nice.
> 
> A few remarks:
> 
> In the settings menu, the rendering of this setting can be a bit nicer. I recommend to describe the setting there in a way different from "Hyp_visibility", but just with a few words, such as "Visibility of hypotheses". Moreover, the list does not get rendered very nicely, as the backticks are just visible. Finally, but this is maybe not so much an issue here, but more for the Waterproof theme, in the Waterproof theme the drop down menu cannot be read.
> 
> I am thinking maybe we can use "We know:" instead of "Context".

Adopted all of this. I move the descriptions of options into the "enumDescriptions", which is more compact anyway.

Created #190 for the theming issue:)

PR #192, opened 2025-07-29
Added flashing buttons for side panel

Alisakondratyev:
### Description
This pull request changes the behavior of the buttons from becoming greyed out when clicked, to flashing when clicked.

### Changes
I have added functionality for the buttons to flash when they are clicked, and a side panel is opened. I have removed the old functionality in which the buttons would become greyed out when side panels were opened.

### Testing this PR
You can test this by opening and closing different side panels, and opening new exercise sheets.


.direnv/bin/nix-direnv-reload
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
pimotte, line 1: Is this file supposed to be in here?

.vscode/settings.json
```diff
@@ -1,3 +1,15 @@
 {
-    "js/ts.implicitProjectConfig.target": "ESNext"
+    "js/ts.implicitProjectConfig.target": "ESNext",
```
pimotte, line 2: This is something that I think the default vscode settings keep generating somehow, let's keep this.

src/webviewManager.ts
```diff
@@ -109,9 +110,16 @@ export class WebviewManager extends EventEmitter {
         webview.on(WebviewEvents.message, (msg: Message) => {
             this.onToolsMessage(name, msg);
         });
-        webview.on(WebviewEvents.change, (state) => {
+        /**webview.on(WebviewEvents.change, (state) => {
```
pimotte, line 113: Let's remove this commented code

src/webviews/sidePanel.ts
```diff
@@ -11,50 +11,42 @@ export function addSidePanel(context: vscode.ExtensionContext, manager: WebviewM
...
-        // If the view is not available, return without updating the styles
+    private flashButton(buttonId: string) {
+        // If the view is not available, return without flashing
         if (!this._view) return;
-
-        // Post a message to the webview to update the button transparency
+        // Post a message to the webview to flash the button
         this._view.webview.postMessage({
-            type: 'trans',
+            type: 'flash',
```
pimotte, line 36: Technically we should add this as a MessageType in `./shared/messages`, but since this message is only internal to this component, I'm not really opposed to leaving this as is. I'd maybe add a TODO to figure out if we can do this without using a webview/messages?

pimotte (changes requested):
Works well! I left some comments mostly related to maintenance.

pimotte (approved):
The post message is indeed internal, so I'm happy to merge this as is. Thanks!

PR #199, opened 2025-08-06
Aligning suggestions

pimotte:
### Description
Improves feedback from Help/Expand.

### Changes
- Both Help and Expand now supply actionable messages to replace/insert that are triggered by specific formatting of messages.
- Expand the definition panel is removed as underlying functionality was removed. 
- Adopt suggestion buttons have been added when the goal window gives a hint.
  - These buttons insert when the cursor is at a single position and replace when there is a selection.

### Testing this PR
Test in conjunction with [this PR on coq-waterproof](https://github.com/impermeable/coq-waterproof/pull/165)

Check the following messages and actions, also check if the actions do the correct thing when indented.
- Unfold (both goal and in hypotheses.
- Help
  - All situations (see Help.v, including when Help. is used in a shielded goal.)
- How to use message after naming an hypothesis.
- Feedback on hypotheses (Use `Waterproof Enable Hypothesis Help`).
  - We claim that
  - Named hypotheses
- Adopting suggestions in the goal window (all cases, see Goals.v for different cases of wrapped goals)


jim-portegies:
In terms of inserting, if you enable `Waterproof Enable Hypothesis Help.` inserting works nicely if you want to insert after the sentence, but if you want to insert something at a different place it becomes cumbersome.

pimotte:
> In terms of inserting, if you enable `Waterproof Enable Hypothesis Help.` inserting works nicely if you want to insert after the sentence, but if you want to insert something at a different place it becomes cumbersome.

It's kind of troublesome to special case for this one, so I think any solution would apply to all messages that go through notice.

We could match the behavior to the adopt suggestion button in the goal window: Insert at cursor or replace selection, depending on if there is a selection.

I'm tempted to leave it as-is, unless more use cases have this issue: The for the other use cases it really makes sense to do stuff around the error, rather than around the cursor.

pimotte:
> I'm also wondering if we still want to provide the copying option by default on messages, or not anymore if we have a good interface. A middle ground may be to remove the word "Copy"?

Adopted.

I did notice now that the messages are shorter, they pop up next to each other. This already is the case on main, it's just more noticable now. I'll see if I can fix this.

DikieDick (changes requested):
I have tested this PR and it works well, but I have some remarks.

There are cases where the next goal (by default) only appears after entering the bullet point, but the suggested snippet (inserted via `adopt suggestion`) in the goals window also includes a bullet, creating invalid syntax. An easy example is 16 in the tutorial where the suggestion for the inductive step appears only after the bullet has been entered manually.

DikieDick:
>  This already is the case on main,

Only outside of input areas, right?

pimotte:
> > This already is the case on main,
> 
> Only outside of input areas, right?

Oh, yeah, maybe? I fixed that in any case:)

pimotte:
> I have tested this PR and it works well, but I have some remarks.
> 
> There are cases where the next goal (by default) only appears after entering the bullet point, but the suggested snippet (inserted via `adopt suggestion`) in the goals window also includes a bullet, creating invalid syntax. An easy example is 16 in the tutorial where the suggestion for the inductive step appears only after the bullet has been entered manually.

Fixed this for the cases where we know that it is the second message. In the case where we can't really know it, the recommended usage pattern is: Select the bullet before clicking the Adopt suggestion button.

PR #204, opened 2025-08-23
New Settings 

Alisakondratyev:
### Description
This is a first draft of the new settings, and I have changed the namings of all the settings to be camelCase and adjusted them accordingly in  `src/helpers/config-helper.ts`.

### Changes
The new categories are:
User Mode, Editor Appearance, Language and Syntax, Diagnostics, Goals and Info Panel, Language Server, Extension

### Testing this PR
I checked the tutorial file and all seemed to still work.


package.json
```diff
@@ -335,7 +335,7 @@
         }
       },
       {
-        "title": "Waterproof Use Case",
+        "title": "Language and Syntax",
         "properties": {
           "waterproof.standardCoqSyntax": {
```
jim-portegies, line 340: Let's change this here, and at several places, to Rocq syntax

package.json
```diff
@@ -457,36 +472,13 @@
...
-        }
-      },
-      {
-        "title": "Editor",
-        "type": "object",
-        "properties": {
+            "default": false,
+            "description": "Disables the checks that happen when launching the extension. During those checks Waterproof checks whether Coq, coq-lsp and coq-waterproof are installed. Use at your own risk!"
+          },
           "waterproof.updateIgnores": {
```
jim-portegies, line 482: Maybe "Update ignores" doesn't way enough

jim-portegies (commented):
Thanks!

I think that it would be good to move "Detailed error mode" outside of "User mode". Perhaps to diagnostics? I'm also not sure if the title "diagnostics" is clear enough.

Maybe another possible sections would be "Editor experience", and Advanced configuration? I think it would be good if "Admit On Bad Qed." moves to one of these categories, same for Unicode Completion.

I am not sure if

DikieDick:
> I am not sure if "language server options" is clear enough for a layperson.

Personally, I would say it is fine. In theory, after running the automatic setup, people don't have to change it again, right?

I also can't really come up with a category name that encompasses all the settings and is 'simpler' at the same time. 

We could do something like "Backend Configuration", but this is maybe too simple for more advanced user?

jim-portegies:
And adding one of the two with parentheses?

jim-portegies (approved):
Looks good now!

PR #208, opened 2025-08-27
Use viewport hints to check selectively

pimotte:
### Description
Communicate to the LSP what part of the document we're viewing so it only computes what is necessary.
Slight decrease in performance for the use case "open document, scroll down relatively slowly"
Increase in responsiveness for basically any other situation.

### Changes
- Start LSP with check_only_on_request
- Uses "coq/viewRange" notifications to signal the viewport
- Disables the spinner in the progress bar
- Updates language in progress bar to be more neutral and not suggest it's working on something
- Query the LSP to render decorations in the view.

### Testing this PR
Open up documents, try edits at the beginning and the end, and anything that has caused problems in the past with respect to progress/decorations.

TODO:

- [ ] Properly disable spinner

### DIFF
--- a/editor/src/waterproof-editor/editor.ts
+++ b/editor/src/waterproof-editor/editor.ts
@@ -353,29 +353,31 @@ export class WaterproofEditor {
 		func(view.state, view.dispatch, view);
 	}
 
-	public handleScroll(scrollY: number, innerHeight: number) {
-		console.log("Enter handleScroll")
+	public handleScroll(innerHeight: number) {
 		if (!this._view) return;
-		// Use the scroll height, window height and posAtCoords to obtain
-		// the current position of the viewport
-		const viewport = {
-			top: scrollY,
-			bottom: scrollY + innerHeight
-		};
-		const posTop = this._view.posAtCoords({left: 30, top: viewport.top});
-		const posAtBottom = this._view.posAtCoords({left: 30, top: viewport.bottom});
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -319,7 +319,7 @@
           },
           "waterproof.showLineNumbersInEditor": {
             "type": "boolean",
-            "default": false,
+            "default": true,
```
DikieDick, line 322: Do we still want to update the code so that it is not necessary to show linenumbers?

DikieDick (commented):
Works really well. 

Remark: Should we notify the user upon changing the `Continuous checking` configuration, as this requires a restart of the server?

editor/src/waterproof-editor/editor.ts
```diff
@@ -343,6 +420,8 @@ export class WaterproofEditor {
 		if (!state) return;
 		const tr = this._view.state.tr.setMeta(CODE_PLUGIN_KEY, msg);
 		this._view.dispatch(tr);
+		// Document progress uses lines to compute the right size of the decorator
+		this.updateDocumentProgress();
```
DikieDick, line 424: Why is this happening here and not in the line number update function?

DikieDick:
@pimotte, it seems that I cannot add you as a reviewer since this is your PR :)

jim-portegies:
See my comments on waterproof-editor's [PR15](https://github.com/impermeable/waterproof-editor/pull/15)

PR #222, opened 2025-09-04
Rely on ocamlfind for version check coq-waterproof, no longer check version or Rocq

jim-portegies:
### Description
Determines the waterproof version using `ocamlfind`, and the coq -> Rocq version using `rocq`.

### Testing this PR
1. Preferably test in conjunction with the latest Windows installer, open a waterproof file, and make sure that no error popups occur and the file gets checked.
2. In a Linux environment, test in an opam switch where some packages are not installed, or have the wrong version, and see if the popup messages are as expected.


pimotte:
Also, this might be a good PR to squeeze in the update to the new installer link, if that is necessary?

PR #227, opened 2025-09-05
Various fixes for the goals panel. 

DikieDick:
Reopens the goals panel after restart and properly closes it on dispose.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -421,6 +421,7 @@ export class Waterproof implements Disposable {
         this.client = this.clientFactory(this.context, clientOptions, WaterproofConfigHelper.configuration);
         return this.client.startWithHandlers(this.webviewManager).then(
             () => {
+                this.webviewManager.open("goals");
                 // show user that LSP is working
                 this.statusBar.update(true);
                 this.clientRunning = true;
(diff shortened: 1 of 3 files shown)


src/webviews/coqWebview.ts
```diff
@@ -60,18 +61,26 @@ export abstract class CoqWebview extends EventEmitter implements Disposable {
     }
 
     public get isOpened() {
+        if (!this._panel) {
+            // FIXME: Sanity check to make sure the state is set to closed.
```
jim-portegies, line 65: Is this fixme still relevant? If so, can it be refined a bit?

src/extension.ts
```diff
@@ -421,6 +421,7 @@ export class Waterproof implements Disposable {
         this.client = this.clientFactory(this.context, clientOptions, WaterproofConfigHelper.configuration);
         return this.client.startWithHandlers(this.webviewManager).then(
             () => {
+                this.webviewManager.open("goals");
```
jim-portegies, line 424: This is great!

jim-portegies (commented):
It seems to work well! I really like the goal panel opening on starting the extension!

One comment on a fixme because I didn't fully understand.

PR #228, opened 2025-09-06
Chore/port update themes

jim-portegies:
### Description
This is a port from the chore/update-themes branch which was merged to main. It is important, because otherwise the walkthrough is not visible, and syntax coloring in the dark mode does not work properly (because Waterproof dark mode used to be registered as a light mode).

### Testing this PR
1. Open a walkthrough and see it is legible
2. Open darkmode and verify there's good contrast between tactic input text and the background.


DikieDick (approved):
Port looks good to me.

Tested in tutorial and inline code and tactic input contrast is both fine.
Walkthrough is also readable in both the light and dark theme.

In addition, we should also port the readable dropdowns to the release branch: ec61af97842af4b6a07902f2b0b39936933597ce

PR #229, opened 2025-09-06
Fix/syntax coloring

jim-portegies:
### Description
Some small changes to the grammar, so that splitting tactics over multiple lines works better.

### Changes
In the `syntax.grammar` file, a lot of tactics included spaces behind some of the tokens, e.g.
```
"It" space+ "holds" space+ "that " TacticInput {TacticLabelAs TacticInput}? TacticEnd |
```
However, it can be nice to put the statement that holds actually on a new line, but then "that" would no longer be colored as it wouldn't be followed by a space. Therefore we changed this to
```
"It" space+ "holds" space+ "that " TacticInput {TacticLabelAs TacticInput}? TacticEnd |
```
Now, splitting the statement over multiple lines is fine.

Note however, that a statement like
```
It holds that∀ x ∈ [0, 5], x ≤ M. (* on purpose no space between 'that' and '∀' *)
```
is also colored, where we may not want this to nudge users in putting some whitespace in between.
However, I could not quickly find a solution for that, so I would prefer to make an issue for that.

On an unrelated note: now we need to manually run lezer-generator, and it generates files that we push to git. Should the generated files be incorporated in the gitignore, and lezer-generator be run automatically before building?

### Testing this PR
1. Open the tutorial, and play around with tactics over multiple lines, seeing that they get colored correctly.


DikieDick:
Ah okay, so then the fix isn't that simple. Let's maybe keep it this way then.

PR #231, opened 2025-09-06
Fix link to uninstaller

jim-portegies:
### Description
The link to the uninstaller was wrong.

### Changes
This PR fixes that.

### Testing this PR
With a previous windows installation, try to run the automatic installer again. Then first the previous installation should get uninstalled.


pimotte (approved):
Clearly correct, tested through release branch

PR #241, opened 2025-09-18
Compute input area status on Idle and add debounce

DikieDick:
### Description
Using the server status notification, compute the input areas as soon as we receive the `Idle` status. Also debounces the computation required for the input area statuses as it requires non-trivial computation that would otherwise happen after every document check performed by coq-lsp.

### Changes
- Adds `CoqServerStatus` notification type.
- Converts the CoqServerStatus` object to `ServerStatus` before sending to the editor.
- Debounces the `computeInputAreaStatus` function by adding a `setTimeout` call.

### Testing this PR
Open the extension and any waterproof file. The difference is most easily observed with `ContinuousChecking` turned **off** (which is the default anyway). Scroll to any red input area, complete the proof and observe that the input area status bar turns green. Important here is that you do not scroll after completing the proof, as this will trigger the computation of the input area statuses.


src/lsp-client/client.ts
```diff
@@ -204,28 +214,48 @@ export function CoqLspClient<T extends ClientConstructor>(Base: T) {
...
+                    ));
+
+                    // forward statuses to corresponding ProseMirror editor
+                    this.webviewManager!.postAndCacheMessage(document, {
+                        type: MessageType.qedStatus,
+                        body: statuses
+                    });
+                } catch (reason) {
+                    if (wasCanceledByServer(reason)) return;  // we've likely already sent new requests
+                    // FIXME: Do we want to throw here?
```
pimotte, line 255: Maybe log instead of throw? If something goes wrong here it's not that problematic, because scrolling will refresh stuff anyway?

pimotte (approved):
Thanks! Works well, one small comment, I'd say this is ready for merge after fixing that:)

PR #246, opened 2025-10-08
Implement Petanque requests

DikieDick:
### Description
Basic implementation of the petanque interface. Supports using petanque to execute commands in isolation.

### Changes
Add petanque types and reworks `executeCommand` to create a petanque state on which to execute the supplied command.

### Testing this PR
Open the extension and use the `Help` button, `ctrl-h` or the context menu to ask for help. The Help panel should open and display the output of the `Help.` command as usual.


pimotte:
> > Is this only a maintanance thing, or does it impact how it looks to the user? If the former, I'd say go for it, the latter, I'd wait.
> 
> Because the message is of type `string` it renders like this: image
> 
> So, I guess it only impacts how it looks for the user.
> 
> Note that we also have twice as many messages as necessary, but that is a separate issue

Just to be sure, the reason we have twice as many messages: We use both the info and notice channel, and with default settings, we only show one in the goals view and the others as popups. (You probably know this, but if you didn't it will save you some debugging)

I agree with Jim: If this is needed for AI support, I'm happy to accept the temporary UI impact.

DikieDick:
I have created #251 to track these two problems

jim-portegies:
When I try to ask for Help from the panel I get:
```
Error: Error when trying to execute command 'Help.': Error: JLang.Point.PointLC.t
```

Moreover, for some reason also in the diagnostics panel I get both info and notice messages, even if my settings say otherwise.

DikieDick:
I spent some time updating it, it now looks like:
image

DikieDick:
The problem with the levels has been addressed in https://github.com/impermeable/coq-waterproof/pull/196

jim-portegies (approved):
It looks good! I think eventually the help messages became much clearer.

PR #250, opened 2025-10-20
Update package.lock and update language ids

DikieDick:
These language ids are the same as the ones used in https://github.com/ejgallego/rocq-lsp/blob/main/editor/code/package.json

Regenerated the package-lock file as it was out-of-sync with the package.json file.

### Testing
Clone this branch, run `npm ci` (this should not complain about package-lock and package.json being out-of-sync).
Test the functionality of the extension by going through some exercises in the tutorial. 

This should also work just fine with coq-lsp 0.2.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,13 +73,21 @@
     "languages": [
       {
         "id": "coq",
+	"aliases": [
+          "rocq",
+          "Coq",
+          "Gallina",
+          "gallina"
+        ],
         "extensions": [
-          ".v",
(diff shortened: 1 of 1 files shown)


pimotte (approved):
If we sync the language ids with rocq-lsp, I think we should at least also include the rocq alias, and perhaps others too. (I have no clue what these ids actually affect)

Tested locally, builds fine, feel free to merge after considering the above.

PR #252, opened 2025-10-23
Expose API from Waterproof extension

DikieDick:
### Description
Uses the vscode extension mechanism to expose an API of the Waterproof extension. Can be used to query for the current goals, help output, etc.

### Changes
One important change is that I add all diagnostics to the diagnostics collection and then process them (see the TODO in the code, we could maybe postpone processing them for waterproof-editor until the checker reached the `Idle` state).

### Testing this PR
Use the extension as you normally would. Nothing should have changed and the overall functionality and feel should be the same.

Pay special attention to the diagnostics in the editor.

### DIFF
--- a/completions/tactics.json
+++ b/completions/tactics.json
@@ -253,22 +253,22 @@
         "boost": 2
     },
     {
-        "label": "Expand the definition of (*name_kw*).",
+        "label": "Expand (*name_kw*).",
         "type": "type",
         "detail": "tactic",
-        "template": "Expand the definition of ${infimum}.${}",
-        "description": "Expands the definition of the keyword (*name_kw*) in relevant statements in the proof, and gives suggestions on how to use them.",
-        "example": "Expand the definition of upper bound.",
+        "template": "Expand ${infimum}.${}",
+        "description": "Uses the definition and alternative characterizations of (*name_kw*) in relevant statements in the proof, and gives suggestions on how to use them.",
+        "example": "Expand infimum.",
         "advanced": false,
         "boost": 2
     },
     {
-        "label": "Expand the definition of (*name_kw*) in ((*expression*)).",
+        "label": "Expand All.",
         "type": "type",
         "detail": "tactic",
-        "template": "Expand the definition of ${infimum} in (${0 = 0}).${}",
-        "description": "Expands the definition of the keyword (*name_kw*) in the statement (*expression*).",
-        "example": "Expand the definition of upper bound in (4 is an upper bound for [0, 3).",
+        "template": "Expand All.${}",
+        "description": "Uses a number of definitions and alternative characterizations to reformulate relevant statements.",
+        "example": "Expand All.",
         "advanced": false,
         "boost": 1
     },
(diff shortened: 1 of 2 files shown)


src/api.ts
```diff
@@ -0,0 +1,14 @@
...
+    currentDocument: () => TextDocument;
+    help: () => Promise<Array<string>>;
+    proofContext: () => { lemma: string, soFar: string };
+    tryProof: (steps: string) => Promise<{finished: boolean, remainingGoals: string[]}>; 
+}
+
+export type Hypothesis = {
+    name: string;
+    content: string;
+}
```
jim-portegies: Let's add a newline

src/lsp-client/commandExecutor.ts
```diff
@@ -44,3 +55,12 @@ export async function executeCommand(client: CoqLspClient, command: string): Pro
...
 }
+
+export async function executeCommandFullOutput(client: CoqLspClient, command: string): Promise<GoalConfig<string> & RunResult<number>> {
+    try {
+        const { goalsRes, runRes } = await executeCommandBase(client, command);
+        return { ...goalsRes, ...runRes };
+    } catch (error) {
+        throw new Error(`Error when trying to execute command '${command}': ${error}`);
+    }
+}
```
jim-portegies: Let's add a newline

src/lsp-client/commandExecutor.ts
```diff
@@ -34,6 +34,17 @@ export async function executeCommand(client: CoqLspClient, command: string): Pro
...
 
+        return {
+            goalsRes, runRes, document
+        };
+    } catch (error) {
+        throw new Error(`Error when trying to execute command '${command}': ${error}`);
+    }
+}
+
+export async function executeCommand(client: CoqLspClient, command: string): Promise<GoalAnswer<string>> {
```
jim-portegies: Do these deserve docstrings?

jim-portegies (commented):
It looks quite good and I like the fact that we have this api functionality now. I have a few small remarks.

I've tried around in the extension, and looked at the diagnostics. Also looked at cases where diagnostics span multiple blocks. It seems to work fine.

jim-portegies:
> I will address the rest of the comments regarding `proofContext` later, as I am not quite sure yet if this is the final form of the function or how it achieves its goal.

Would you already want to merge nonetheless, or wait?

src/extension.ts
```diff
@@ -268,6 +266,124 @@ export class Waterproof implements Disposable {
...
+
+        const pos = document.offsetAt(position);
+
+        // This regex is used to find the statement the user is trying to proof.
+        // see: https://rocq-prover.org/doc/V9.0.0/refman/language/core/definitions.html#assertions-and-proofs
+        // We match for one of the `thm_token` followed by `ident_decl` (of which we ignore the universe part (univ_decl))
+        // up to one of {Qed, Admitted, Defined} followed by a dot.
+        // Note that the ident is allowed to contain unicode so this regex operators with the unicode flag enabled (/u).
+        // \p{L} matches all the unicode symbols in the 'letter' category: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape
+        const regex = /(?:Theorem|Lemma|Fact|Remark|Corollary|Proposition|Property)\s+([A-Za-z_\p{L}][A-Za-z0-9_\p{L}']*)\s+[^]*?(?:Qed|Admitted|Defined)\./gu;
```
jim-portegies: I was thinking, if we use this regex, would it make sense to require a space or newline after the final period? I think that would make it more robust.

PR #255, opened 2025-10-27
Update version to pre-release 0.9.24

DikieDick:
### Description
Update version tag to next pre-release that includes the api for use with waterproof-river.

### Changes
Update version to 0.9.24

Also removes `vscode-extension-tester` as we don't use it and its dependencies depend on very old and deprecated packages that mess with the vsce package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "waterproof",
-  "version": "0.9.25",
+  "version": "0.9.24",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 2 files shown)


DikieDick:
> Silly question, but are we skipping 0.9.24?

Ah, I wanted to release this as a pre-release still. But if we feel like making an actual release then I will change it to 0.9.24.

DikieDick:
Woops, I knew something was not right. I should maybe not be doing this when getting tired :)

PR #273, opened 2025-11-14
Bump waterproof-editor to 0.12.0 (strict)

DikieDick:


pimotte (approved):
Tested autocompletion and hover behaviour, I think this is good to go.

PR #283, opened 2025-12-15
Remodellingpanels

raulTUe:
### Description
The help button now appears in the goals panel, below the goals. This reduces the user's cognitive load, letting them get extra help without the need to open a different panel.

### Changes
The help panel has been removed as the help button is now in the goals panel, which is opened by default. The error text has been restricted, as it would clog up the goals panel and the user would have to scroll to get the help button. 

### Testing this PR
Test the separate functionalities (goals and help) to see that they both work as intended. Begin typing in an input cell to see how the error message is restricted

To note: the commit history has become a bit entangled due to multiple merges with main. However, the changes are the final version and accurate.


src/webviews/goalviews/goalsPanel.ts
```diff
@@ -25,4 +28,17 @@ export class GoalsPanel extends GoalsBase {
         super.updateGoals(goals);
     }
 
+    setResults(results: string[]): void {
+        if (results[0] == "createHelp") {
```
pimotte, line 32: This looks like we're abusing a message somewhere, I'd be in favour of updating a message type or including a seperate message for this. (I'm aware this is existing behaviour, but I think this would be a good opportunity to improve it).

pimotte (changes requested):
When testing the help functionality, I get an error when I press the "Help" button:

```
Error: Error when trying to execute command 'Help.': 
Error: Error when trying to execute command 'Help.': 
Error: JLang.Point.t.offset
```

I have a hunch this might be due to coq/rocq-lsp version, but I can't get a functional one even when using an old switch.

pimotte:
I updated my coq-lsp to 0.2.4 and now it works:) I think this is a great improvement over what we used to have, one minor remark still:


Long messages kind of run out of the bar. I'm not sure if we have a reasonable way to allow these to line-wrap?



image

pimotte (approved):
Awesome, looks great!

PR #284, opened 2025-12-30
Adapt to waterproof-editor#custom-menu-bar-entries

DikieDick:
Targets the 'mapping' branch

Language Package can probably be developed separately.


DikieDick:
I still need to update some dependencies across both repositories and incorporate the language packages correctly. 

In particular, the lean language package still needs to be set up correctly.

package.json
```diff
@@ -532,7 +532,14 @@
     "typescript-eslint": "^8.20.0"
   },
   "dependencies": {
-    "@impermeable/waterproof-editor": "0.13.0",
```
pimotte, line 535: It doesn't seem correct to remove this dependency? We're not getting it through the language packs, are we?

package.json
```diff
@@ -532,7 +532,14 @@
...
+    "@codemirror/autocomplete": "6.18.1",
+    "@codemirror/commands": "6.6.2",
+    "@codemirror/language": "6.12.1",
+    "@codemirror/lint": "6.9.2",
+    "@codemirror/state": "~6.5.2",
+    "@codemirror/view": "6.38.8",
+    "@impermeable/codemirror-lang-verbose": "github:impermeable/codemirror-lang-verbose",
```
pimotte, line 541: Do we want to release these on npm before merging this to main? (Or even to mapping?)

pimotte (changes requested):
I tried to review this, but it seems in some broken state. The editor does not load at all and I get `Unrecognized extension value in extension set ([object Object]). This sometimes happens because multiple instances of @codemirror/state are loaded, breaking instanceof checks.` in the browser console.

(This is with having `@impermeable/waterproof-editor` linked, but the fact that it's not in the package.json also seems wrong to me)

package.json
DikieDick: I don't really have an opinion on this. My thought is that they are maybe a bit too specialized for release on npm and installing from GitHub in this way works fine.

package.json
DikieDick: Oh, thanks for catching this!

DikieDick:
> I tried to review this, but it seems in some broken state. The editor does not load at all and I get `Unrecognized extension value in extension set ([object Object]). This sometimes happens because multiple instances of @codemirror/state are loaded, breaking instanceof checks.` in the browser console.
> 
> (This is with having `@impermeable/waterproof-editor` linked, but the fact that it's not in the package.json also seems wrong to me)

The state is not broken, `@codemirror/state` is just really finicky.

The error you are getting can probably be solved by deleting all dependencies in waterproof-vscode and waterproof-editor, then installing again with `npm i` and finally linking waterproof-editor again.

package.json
pimotte: I'm happy to keep this as-is for now and reconsider if it turns out there are drawbacks to doing this.

pimotte (approved):

PR #289, opened 2026-02-05
Mapping

DikieDick:


pimotte:
@DikieDick I've done a little bit of cleanup, there's two things that are known issues to me:

- The cypress tests with mocks are broken
- The typecheck doesn't pass due to dependency issues. Especially this one we should address before releasing waterproof-editor

pimotte (approved):
We decided to incrementally fix the lacking test coverage whenever we touch the relevant code again.

PR #294, opened 2026-02-09
Lean documentation

### LINKED ISSUE #293: doc: Elan instructions

pimotte:
Documents the Lean installation step.

I chose to do this in the place where the Rocq stuff is also mentioned, but this README does fulfill a double role, also being for end-users. My proposal is that we re-evaluate the documentation after we look at the Lean installation story.

Fixes #293

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ The Waterproof vscode extension helps students learn how to write mathematical p
     c. [Manual installation on Windows with installer](#manual-installation-on-windows-with-installer)
 
     d. [Manual installation on Windows with WSL](#manual-installation-on-windows-with-wsl)
-3. [Getting Started](#getting-started)
+3. [Optional dependencies for (Verbose) Lean](#optional-dependencies)
+4. [Getting Started](#getting-started)
 
 # Automatic installation instructions for Windows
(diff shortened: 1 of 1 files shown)


DikieDick (commented):
Looks good, but shouldn't we also mention that the Waterproof genre is required?

At the moment I would not consider Lean Verbose optional though, as all our completions are based on it.

pimotte:
This part of the documentation is meant for users (students), as I see it, and those don't need to worry about the genre or verbose.

The way I'm phrasing it here is that I want to suggest that both Lean and Verbose Lean are optional, which matches the current use in practice.

DikieDick:
I see, with your comment it makes more sense.

Still I think we should include or clarify that (Lean+Waterproof-genre+Verbose) is optional.

What I mean to say is that Lean in and of itself is not optional, as just having a plain .lean file (without our specific genre) will just crash.

pimotte:
Shaved the language a bit.

I think anything more really belongs in dev or teacher documentation. (The latter category we don't even have, but we're getting very close to needing it.)

DikieDick (approved):

PR #296, opened 2026-02-10
Fix exact replacement

### LINKED ISSUE #290: [Chore] Fix exact? insertion to use existing API
Currently using exact? in a lean file allows inserting the suggestion by clicking the infoview, but the way this is implemented is kind of hacky. We should be using the correct method.

@DikieDick Could you add a reference to where this is in the code?

Tammo0987:
### Description
This fixes applying suggestions. The main problem it solves, that instead of working directly on the vs code document layer, we now take the replace message from the lsp and send it through the prose mirror. Instead of using the already existent insert message, I introduced a new one for replacing, to exactly use the information from the lsp. Using the insert message felt to hacky.

There are also changes in the waterproof-editor repo (https://github.com/impermeable/waterproof-editor/pull/56)

Closes #290 

### Changes
- New replace text message
- Reimplementation of applyEdit

### Testing this PR
Use a lean waterproof document and type in exact? (or apply?, should work the same) and click on one of the suggestion in the info panel.

### DIFF
--- a/src/infoview.ts
+++ b/src/infoview.ts
@@ -279,21 +279,8 @@ export class InfoProvider implements Disposable {
             await env.clipboard.writeText(text)
         },
 
-        insertText: async (text, _kind, _tdpp) => {
-            wpl.log(`[InfoProvider] insertText called with: ${text}`);
-            const document = this.client.activeDocument;
-            if (!document || !this.client.webviewManager) {
-                wpl.log(`[InfoProvider] Cannot insert text: no active document or webviewManager`);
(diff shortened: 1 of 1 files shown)


pimotte (approved):
Looks good to me, tested manually.

Before merging, I just want to check with @DikieDick: I recall you mentioning that we could reuse the insert message for this, but that uses the cursor position, and this has the positions provided from the Lean Widget, so I think adding the new message is indeed the way to go, agree?

PR #299, opened 2026-02-11
Decouple Lean and Rocq starting

### LINKED ISSUE #292: Waterproof does not start when either language server fails to start
This is annoying when Waterproof is used for either Rocq or Lean.

Example: open .mv file without having lake installed -> lake not found -> can't start server -> waterproof checker does not start.

pimotte:
### Description
This decouples Lean and Rocq startup, and refactors the skipLaunchChecks setting.

### Changes
We will try to start both the Lean and Rocq stacks. The active checkers are displayed in the checker status.

The skipLaunchChecks no makes it possible to skip the checks for a particular language and then start that server anyway. This allows the Lean use case to include a setting to just launch only Lean in a folder with exercises.

The current user experience is friendly for Rocq users. Lean users that have not configure skipLaunchChecks will get the installation instructions popup for Rocq.

### Testing this PR
Start the editor without Lean installed or with a broken Rocq installation (e.g. change the Waterproof path to something that's wrong), and observe that only "lean4" or "rocq" is displayed in the checker status.
If both are available, it should show both.

Fixes #292

### DIFF
--- a/__tests__/prosedoc-construction/top-level-construction.test.ts
+++ b/__tests__/prosedoc-construction/top-level-construction.test.ts
@@ -47,6 +47,7 @@ $$\`x^2 + y = z\`
 A list:
   1. *Italicized* text
   2. $\`y = z - x^2\`
+  3. \`Inline code\`
 `;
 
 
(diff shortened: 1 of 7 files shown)


DikieDick:
Ah, maybe we should also change `waterproof.path` into something like `waterproof.rocqLspPath`?

DikieDick:
> I am not strongly oppposed to changing waterproof.path, but I'm also not convinced it's worth breaking backwards compatibility. If we do it, I think we should also update some of the Rocq/coq language and do it in a separate PR.

Okay, let's keep it the way it is then.

PR #300, opened 2026-02-16
Parse inline math as single entity to prevernt conflict with inline literal code `...`

pimotte:
### Description
Simplifies parsing of Lean documents to prevent crashing on `...`

### Changes
Parse inline math as a single element.

### Testing this PR
Open a Lean document with inline code and observe it renders.


DikieDick:
I fixed the math inline. This is fine to merge now imo.

PR #301, opened 2026-02-16
Hide hypotheses in lean infoview

### LINKED ISSUE #291: Ensure Lean Infoview respects the hypothesis display options.
This should be achieved using the `lean4.infoview.style` option (per [#lean4 > Pre-RFC: Hide all hypotheses in infoview](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Pre-RFC.3A.20Hide.20all.20hypotheses.20in.20infoview/with/572183188))

Michaillus:
### Description
~~This hides all the hypotheses in lean infoview, showing only the goal. I believe it isn't possible to use lean4.infoview.style as suggested in the issue, since waterproof extension doesn't use the same settings. Instead, I added a CSS style to the HTML code where infoview is created. Hypotheses in HTML are just divs without specific class, so I couldn't hide them directly. However, the goal div after them has a property data-is-goal="true". I hide all the siblings before goal div using this property.~~

- This makes waterproof lean extension to respect waterproof.visibilityOfHypotheses setting.
This was eventually achieved in a different way from proposed CSS modification to allow more flexibility. I modify rpc responses from lean language server, which contain hypotheses, before they are sent to the infoview. The setting options are implemented as follows:
"all" - all hypotheses are kept
"none" - all hypotheses are removed
"limited" - hypotheses with type in the response is marked as "text" are removed. Those are hypotheses with types like N and R, but also simple types like P, where P had type Prop, or Prop itself. Compound type like a > 1 or P ^ Q are kept. This was the only meaningful way to filter the hypotheses that I found using response structure.
- The setting change is applied instantly when something changes in the infoview.
- All hypotheses are still shown in error message. The reason is that rpc diagnostics responses don't have hypotheses in a separate field, so it's not possible to filter them.

Closes #291 

### Changes

- Hypotheses list inside responses from lean language server is filtered depending on visibility setting.

### Testing this PR

I added as a comment examples that I used to check the feature. Try turning on different hypotheses visibility options on those.


pimotte (changes requested):
As discussed in the meeting, we want this to respect our option, I've outlined this in the issue: #291

I don't think this PR is blocking for merging the mapping branch, so if it comes after the merge, that's fine.

Michaillus:
Examples for testing ( code format fixed)

````

import WaterproofGenre
import Verbose.English.All
open Verbose English

## EXAMPLE

```lean
theorem hypothesis_test (n : ℕ) (L : List ℤ) : True := by
```
:::input
```lean
-- Logical Propositions (using 'have')
  have h1 : n > 0 := sorry
  have h2 : ∀ x ∈ L, x ≠ 0 := sorry
  
  -- Local Data Definitions (using 'let')
  let m := n + 5
  let s := "Hello, Waterproof"
  
  -- Complex Mathematical Objects
  let f : ℝ → ℝ := fun x => x^2 + 1
  have h3 : f 2 = 5 := by sorry
  
  -- An existential statement (Prop)
  have h4 : ∃ k : ℕ, k > n := ⟨n + 1, by exact Nat.lt_succ_self n⟩

  -- Creating grouped hypotheses (sharing a type)
  let ⟨(a : ℤ), (b : ℤ), (c : ℤ)⟩ : ℤ × ℤ × ℤ := ⟨1, 2, 3⟩

  trivial
```
:::

```lean
QED
```

## TWO GOALS EXAMPLE

```lean
theorem and_split_test (P Q : Prop) (hp : P) (hq : Q) : P ∧ Q := by
```
:::input
```lean
-- Logical Propositions (using 'have')
  have n : ℤ := 3
  have L : List ℤ := sorry
  have h1 : n > 0 := sorry
  have h2 : ∀ x ∈ L, x ≠ 0 := sorry
  
  -- Local Data Definitions (using 'let')
  let m := n + 5
  let s := "Hello, Waterproof"
  
  -- Complex Mathematical Objects
  let f : ℝ → ℝ := fun x => x^2 + 1

  constructor
  -- Goal 1: ⊢ P
  -- Goal 2: ⊢ Q
  · exact hp
  · exact hq
```
:::

```lean
QED
```

````

pimotte (commented):
Clean solution! I took the liberty of pushing a small commit updating the wording of the setting to be language-agnostic.

P.S. Note that your example mixes regular Lean with Verbose Lean. If you do Verbose Lean you need the Given: Assume: structure along with the QED. In addition, to make it easily copy-able, you can use a quadruple ` on the outside and then paste the whole file, which presents it all in one codeblock.

pimotte (approved):

pimotte (changes requested):
Could you look into unit testing the filter methods? I'd be okay with increasing visibility from private in order to do so.

(I originally approved this, thinking we could go without, but there's some implicit design decisions in here that I think make it worthwhile to do that anyway.)

pimotte (approved):
Excellent, looks good, thanks!

Issue #149, opened 2025-02-18
Side Panel Bugs


Alisakondratyev:
### What happened?
After closing the goals panel and then typing in an input area, clicking the goals button won't open the goals panel.

### How can we reproduce this issue?
1. Run the extension with an exercise sheet open, the goals panel will be opened at initialization. 
2. Close the goals panel, and then click on any input area in the exercise sheet and type (or delete) a character.
3. Click the goals button, the goals button will flash but the goals panel will not open.

### What was supposed to happen?
The ideal behavoir is that upon opening an exercise sheet the goals button is greyed out and the goals side panel is opened. All other buttons are blue.

### Additional Information
Please specify the operating system and the version of the Waterproof extension you were using:
#### Operating System: 
- Windows 11

#### Waterproof Version: 
0.10.1, 3.1.0+9.0

### What happened?
5 main bugs were found. 
The first 4 are when an exercise sheet is opened when the extension is run.
1) Greying out of goals, logbook, and debug buttons upon clicking into the exercise sheet that is opened upon running the extension. Only the goals side panel is opened.
2) All buttons become blue when a new exercise sheet is opened, regardless of if the side panels are opened.
3) In this new sheet the side panel buttons will become greyed out only when a side panel that is not yet opened is clicked on. In that case, only the side panel button that is clicked in the new sheet and the last to be opened side panel will have their buttons greyed out.
4) Clicking on the waterproof checker button at the bottom left corner of the screen will result in all of the currently opened side panel buttons being greyed out in addition to the logbook and debug buttons to be greyed out. The goals side panel will close.
The fifth occurs when the extension is run after previously all the exercise sheets are closed.
5) Once an exercise sheet is opened only the logbook button is greyed out, and the goals side panel is opened, but the goals button is not greyed out.

### How can we reproduce this issue?
An example of reproducing each issue is given.
1) Run the extension, click into the exercise sheet that is opened once the extension is started. The goals, logbook, and debug buttons will be greyed out, and only the goals side panel will be opened.
2) Go through same steps as behavior 1, and now click on the help button and then open a new exercise sheet. Click into a cell in the exercise sheet and all of the buttons will be blue, but the goals and help side panel will be opened. 
3) Go through same steps as behavior 2. Try to click on the goals and side panel buttons, there should be no change to the button colors. Then click on the expand definition button and the help and expand definition button will become greyed out (not the goals button though) and the expand definition side panel will open.
[...]

PR #308, opened 2026-02-23
Fix for Goals Panel not re-opening, as per Issue #149

raulTUe:
### Description
Fix for Issue #149, which caused the goals panel to not opening after a cursor changes occurred while it was closed.
### Changes
Before setting the webview state of the (goals) panel to visible, it first checks that this panel exists. This ensures that the panel is not mistaken for being visible when it isn't.
### Testing this PR
Use the same steps to reproduce Issue #149, the goals panel should open reliably instead.


pimotte (approved):
Works cleanly!

PR #309, opened 2026-02-23
Busy indicator

GitLuckier:
### Add busy indicator support to the VS Code extension

 **Depends on the `mapping` branch of the editor repo being merged simultaneously.**

Wires up the LSP `$/coq/executionInformation` notification so the editor can display a busy indicator on the line currently being checked by coq-lsp, and adds a setting to opt out.

**What changed**
- `requestTypes.ts` defines `ExecutionInformationParams` and the notification type for `$/coq/executionInformation`
- `client.ts` subscribes to the notification, converts the LSP range to document offsets, and forwards a `MessageType.executionInfo` message to the webview
- `types.ts` adds `send_execinfo` to `CoqLspServerConfig`, populated from the new setting
- `config-helper.ts` adds `WaterproofSetting.SendExecInfo` mapped to `"sendExecInfo"`
- `Messages.ts` adds the `executionInfo` message type with a `{ from, to }` body, and moves `flash` into alphabetical order
- `index.ts` handles `executionInfo` by calling `setBusyIndicator(range.from)`, and clears indicators when a `progress` message arrives with an empty progress array
- `package.json` registers `waterproof.sendExecInfo` (boolean, default `true`)
- `messages.md` documents the new message

### DIFF
--- a/developer-resources/messages.md
+++ b/developer-resources/messages.md
@@ -97,6 +97,18 @@ This message does not have a body.
 errorGoals
 ```
 
+### `executionInfo`
+#### Description
+Sent by the extension to the editor to indicate the range of the sentence currently being checked by the language server. Used to display a busy indicator at the relevant position in the editor.
+
+#### Body
+```ts
+{
+    from: number, // Start offset of the sentence being executed
+    to: number    // End offset of the sentence being executed
+}
+```
+
 ### `init`
 
 #### Description
--- a/editor/src/index.ts
+++ b/editor/src/index.ts
@@ -186,7 +186,14 @@ window.onload = () => {
 			case MessageType.progress:
 				{
 					const progressParams = msg.body;
-					editor.updateProgressBar(progressParams);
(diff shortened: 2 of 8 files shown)


pimotte:
> @GitLuckier I love this animation! :)
> 
> One very preliminary suggestion that I have is to make the gutters opaque as currently scrolling the content into the gutter looks kinda strange: image

Agree with the suggestion, but this is an existing issue. I created #311 for this. @GitLuckier If you feel this makes sense to include in this PR, go ahead, otherwise it's a nice issue to pick up later.

GitLuckier:
I made some final changes. I made the gutter transparent (this looks better in some cases, namely when the text exceeds the line length. Note that this does not fix #311, this is the gutter on the right side) and I added fallback units in case  `lh` isn't supported. 

I think #311 should be separate issue and not included in this PR. 

I will now unmark this PR as draft.

pimotte (approved):

PR #312, opened 2026-02-25
Extend logic in determineProofStatus to check for cases where the QED is a 'term'.

DikieDick:
For example, in a proof of the form
```lean
Fix b >
QED
```
The input area is incorrectly colour green since the QED acts (incorrectly) as input to the Fix tactic. Fortunately, this throws an error as QED is not a valid term. We use this error to check for these cases and mark the input area incomplete.

### Testing

Play around with some lean files. Correct proofs should still be coloured green and incomplete proofs red.
Try any of the corner cases where QED becomes input to the tactic as described above and notice that the input area is not coloured green (permanently).


pimotte (approved):
LGTM.

While testing, I did notice a lot of flickering, but this is an existing issue. I created https://github.com/impermeable/waterproof-vscode/issues/313 for this.

Issue #319, opened 2026-03-04
[Bug] [Lean] [Spike] Verified file up to line is inconsistent


pimotte:
### What happened?

When opening files from [this repo](https://github.com/impermeable/bewijzen-in-de-wiskunde-lean/) (solutions secret for students)
the progress bar is quite inconsistent. It will either say "up to line 1", "up to line x", where x is the line before `#doc` or not be present, even when goals after line 1/x are already shown.

### How can we reproduce this issue?

Example:
1. Open any worksheet from the linked repo
2. Observe the inconsistent behaviour

### What was supposed to happen?

The progress bar should reflect the checking progress. Same for the progress marker.

### Additional Information

The reason I've marked this as a spike is because I'm not sure it has a reasonable resolution: It could very well be that the LSP indeed only communicates this information with the given granularity.

When investigating this, pay attention to the proof bars as well.

Outcomes of this spike are answers to the following
- Is the problem indeed the LSP information not being granular enough, or is there another bug?
   ([#verso > Fine grained file progress](https://leanprover.zulipchat.com/#narrow/channel/576452-verso/topic/Fine.20grained.20file.20progress/with/578085116) suggests that the information should be available somewhere)
- If yes, Could we get more finegrained progress from the LSP?
  - If yes, describe the technical steps needed to improve this
  - If no, describe the best mitigation possible to retain this information
- Design proposal or prototype for the progress reporting behaviour. 

Also possibly related to https://github.com/impermeable/waterproof-editor/issues/77

pimotte:
Hypothesis: From the Lean extension we get the information both that the `#doc` is elaborated, but also where we are in proofs. We currently just look at the first one. We might want to do something along the lines of "look at bottom

@GitLuckier I've taken the liberty of assigning this to you, since it seems a good follow-up issue to what you have been doing so far.

pimotte:
Fixed by #336

PR #320, opened 2026-03-08
Busy indicator lean

GitLuckier:
### Description
Added messaging support on the Lean side to send file progress to the editor. 

### Changes
The most important change is that in `onFileProgress` in the Lean client. Aside from a few type changes, nothing of note. The busy-indicator was always there, just never called until now.

### Testing this PR
You can copy+paste this code in your favourite .lean file:
```lean
#eval do  
  IO.sleep 10000
 ``` 
After 0.5 seconds, the busy-indicator should appear then disappear after 9.5 seconds. 

Closes issue number 77 in the editor repo


pimotte:
Looks good! I'll merge this, but it does suffer from #319

In exercise sheets with the Waterproof genre, the indicator is placed at the #doc command, not at the commands that are actually taking a long time.

Once we fix #319, I'm fairly certain the behavior will be correct.

PR #327, opened 2026-03-17
Fix proof bar turning green prematurely in Lean (closes #313, #316)

### LINKED ISSUE #313: [Bug] [Lean] Proof status bar flickers between green and red while typing
### What happened?
When typing in the Lean mode, the status bar flickers between red and green while typing tactics.

Example:
1. Open the ATP test file
2. Type "We apply" one character at a time
3. Observe the bar flash green briefly before returning to red.

The bar should not be flickering.

Considerations for implementation:
- Debouncing the change might seem a solution that solves this issue, but also seems unstable: A smart choice of defaults or not changing the colours in absence of LSP messages is preferred.

### LINKED ISSUE #316: [Bug] [Lean] Last input area in Lean document does not get marked as green
### What happened?
The last input area of a Lean document is marked yellow instead of green if the proof is complete. See [the exercise sheets here](https://github.com/impermeable/bewijzen-in-de-wiskunde-lean) for examples. (Note that the exercise solutions in there should remain secret, don't share those publically)

Example:
1. Open sheet1_conjunction.lean
2. Wait for Lean to check the document
3. Observe the yellow bar

The last bar should be green too.

A solution is possibly along the lines of #312, since what is special is a "Victory 🎉 " message, and this might throw off our check.

GitLuckier:
### Description

The proof status bar for Lean input areas was turning green at inappropriate times, most notably while the user was still typing (e.g. after typing `"We "`), or when a proof contained invalid syntax (like "Help\n• Fix a : ℝ\n") that happened to produce empty goals. This PR rewrites `determineProofStatus` in the Lean client to correctly reflect proof state by (1) guarding against Lean still processing, (2) checking for error-level diagnostics, and (3) only then querying goals.

The race condition that I mentioned in the issue description has been resolved.

As a bonus, it also fixes #316, where the last input area of a Lean document was incorrectly marked yellow instead of green. This was caused by the following lines in `determineProofStatus`:

```typescript
        const content = document.getText();

        const nextQed = content.indexOf("\nQed\n", document.offsetAt(inputArea.start));
        const nextProof = content.indexOf("\nProof:\n", document.offsetAt(inputArea.start));

        if (nextProof && nextQed >= nextProof) {
            return InputAreaStatus.Invalid;
        }
```

This bug was messing with my unit-tests, that is why I deleted it in this PR. The fact that `\nQed\n` is not properly capitalised, indicates to me that this is dead code that can safely be removed.

### Changes

**`src/lsp-client/client.ts` - `onDocumentChanged` abstract hook**

An abstract `onDocumentChanged(): void` method is added to `LspClient`. It is called whenever a text-document change is detected. This replaces the previous approach of setting a `protected isBusy` flag directly in the base class, giving each subclass full control over how it responds to document changes.

**`src/lsp-client/lean/client.ts` - `isBusy` flag and rewritten `determineProofStatus`**

A `private isBusy: boolean` field (initially `true`) is added to `LeanLspClient`. It is set back to `true` in `onDocumentChanged`, and set to `false` once Lean reports no remaining processing ranges in `onFileProgress`.

The new logic in `determineProofStatus` is:

1. **If `isBusy` is `true`**, return `Incorrect` immediately. Lean hasn't finished processing the input area yet, so an empty goals response would mean "not checked yet" rather than "proof complete".
2. **If there are any `Error`-severity diagnostics whose range intersects the input area**, return `Incorrect`. This catches cases where an incomplete tactic keyword (e.g. just `"We"`) causes Lean to report empty goals at the query position despite the proof being invalid.
3. **Otherwise**, request goals at the end of the input area. Return `Correct` if goals are empty, `Incorrect` if not.

**`src/lsp-client/lean/client.ts` - `onFileProgress` ordering**

`super.onFileProgress(progress)` is now called *before* processing the progress ranges, so that LSP ranges are converted to VS Code `Range` objects before the Lean client reads them.

**`src/lsp-client/rocq/client.ts` - `onDocumentChanged` no-op**

The Rocq client implements `onDocumentChanged` as a no-op, since it tracks its own busy state via the `coqServerStatus` notification.

**`__tests__/lsp-client/lean/client.determineProofStatus.test.ts` - new test file**

Seven unit tests cover the key branches of `determineProofStatus`.

### Testing this PR

**Automated tests:** Run `jest __tests__/lsp-client/lean/client.determineProofStatus.test.ts`. All seven tests should pass.

**Manual testing (for #313 - premature green bar):**
1. Open a Lean exercise sheet.
2. Inside an input area, type a partial tactic such as `We`, `We ap` or `We apply` and pause briefly.
3. The bar should remain red, it must **not** flash green while Lean is still processing or while the tactic is incomplete.
4. When typing invalid lean syntax (e.g. `Help\n• Fix a : ℝ\n`), the bar shouldn't turn green like it used to.
5. Valid proofs should still make the bar green.

**Manual testing (for #316 - last bar stuck yellow):**
1. Open a Lean file with multiple input areas (e.g. `sheet1_conjunction.lean`).
2. Complete all proofs correctly and wait for Lean to finish checking.
3. The **last** input area's bar should turn green, not yellow.

**Known (non-)issues:** Typing `hint` or `sorry` inside an input area still causes the bar to turn green. The `sorry` case may be intentional (though we could make it return `Invalid` or something); the `hint` case was pre-existing and is outside the scope of this PR.

closes #313, #316

### DIFF
--- a/__tests__/lsp-client/lean/client.determineProofStatus.test.ts
+++ b/__tests__/lsp-client/lean/client.determineProofStatus.test.ts
@@ -183,4 +183,16 @@ describe("LeanLspClient.determineProofStatus", () => {
         expect(await call(instance)).toBe(InputAreaStatus.Correct);
     });
 
+    it("returns Incorrect when requestGoals resolves with null", async () => {
+        const instance = makeClient();
+        jest.spyOn(instance, "requestGoals" as any).mockResolvedValue(null);
+        expect(await call(instance)).toBe(InputAreaStatus.Incorrect);
+    });
+
+    it("returns Incorrect when requestGoals resolves with undefined", async () => {
+        const instance = makeClient();
+        jest.spyOn(instance, "requestGoals" as any).mockResolvedValue(undefined);
+        expect(await call(instance)).toBe(InputAreaStatus.Incorrect);
+    });
+
 });
--- a/__tests__/lsp-client/lean/client.determineProofStatus.test.ts
+++ b/__tests__/lsp-client/lean/client.determineProofStatus.test.ts
@@ -1,0 +1,186 @@
+jest.mock("vscode", () => {
+    const Position = class {
+        constructor(public line: number, public character: number) {}
+        translate(lineDelta: number, charDelta: number) {
+            return new Position(this.line + lineDelta, this.character + charDelta);
+        }
+    };
+
+    const Range = class {
+        constructor(
+            public start: InstanceType<typeof Position>,
+            public end:   InstanceType<typeof Position>
+        ) {}
+        intersection(other: InstanceType<typeof Range>): InstanceType<typeof Range> | undefined {
+            const startLine = Math.max(this.start.line, other.start.line);
+            const endLine   = Math.min(this.end.line,   other.end.line);
+            if (startLine > endLine) return undefined;
+            return new Range(new Position(startLine, 0), new Position(endLine, 0));
+        }
+        get isEmpty() {
+            return this.start.line === this.end.line
+                && this.start.character === this.end.character;
+        }
+    };
+
+    return {
+        Position,
+        Range,
+        DiagnosticSeverity: { Error: 0, Warning: 1, Information: 2, Hint: 3 },
+        EventEmitter: class {
+            fire()  {}
+            event = () => ({ dispose: () => {} });
+        },
+        workspace: {
+            getConfiguration:         jest.fn(() => ({ get: jest.fn((_key: string, def?: any) => def) })),
+            onDidChangeConfiguration: jest.fn(() => ({ dispose: jest.fn() })),
+            onDidChangeTextDocument:  jest.fn(() => ({ dispose: jest.fn() })),
+        },
+        languages: {
+            createDiagnosticCollection: jest.fn(() => ({ set: jest.fn(), dispose: jest.fn() })),
+            getDiagnostics:             jest.fn(() => []),
+            onDidChangeDiagnostics:     jest.fn(() => ({ dispose: jest.fn() })),
+        },
+        window: {
+            createOutputChannel: jest.fn(() => ({
+                appendLine: jest.fn(),
+                dispose:    jest.fn(),
+            })),
+        },
+    };
+}, { virtual: true });
+
+jest.mock("vscode-languageclient", () => ({
+    LogTraceNotification:  { type: "$/logTrace" },
+    RequestType:           jest.fn().mockImplementation(() => ({})),
+    NotificationType:      jest.fn().mockImplementation(() => ({})),
+    DocumentSymbolRequest: { type: {} },
+}), { virtual: true });
+
+jest.mock("vscode-languageserver-types", () => ({
+    VersionedTextDocumentIdentifier: { create: jest.fn((uri, v) => ({ uri, version: v })) },
+}), { virtual: true });
+
+jest.mock("../../../shared",        () => ({ MessageType: {}, FileProgressKind: { Processing: 1 } }), { virtual: true });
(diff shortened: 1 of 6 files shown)


__tests__/lsp-client/lean/client.determineProofStatus.test.ts
```diff
@@ -0,0 +1,186 @@
+jest.mock("vscode", () => {
```
pimotte, line 1: This is quite an extensive mock setup that might be useful for other tests? I think we should keep an eye on it and extract it later if we can generalize it.

src/lsp-client/lean/client.ts
```diff
@@ -96,9 +105,9 @@ export class LeanLspClient extends LspClient<LeanGoalRequest, LeanGoalAnswer> {
                     body: { from, to },
                 });
             }
+            this.isBusy = processingRanges.length > 0;
```
pimotte: This line breaks for me. processingRanges is not defined anywhere?

pimotte (changes requested):
For me all bars remain blue and the unit tests fail on an undefined variable, so you may have forgotten to commit a file?

src/lsp-client/lean/client.ts
GitLuckier: Oh sorry, some changes from another issue accidentality got mixed in, I'm on it

src/lsp-client/lean/client.ts
```diff
@@ -161,33 +172,42 @@ export class LeanLspClient extends LspClient<LeanGoalRequest, LeanGoalAnswer> {
...
-        const nextProof = content.indexOf("\nProof:\n", document.offsetAt(inputArea.start));
-
-        if (nextProof && nextQed >= nextProof) {
-            return InputAreaStatus.Invalid;
+        // If Lean hasn't finished processing up to the end of this input area, an empty goals
+        // response simply means "not checked yet" rather than "proof complete".  Guard against
+        // that to prevent the bar from turning green prematurely (e.g. while typing "We ap" or
+        // when the proof has invalid syntax like "Help\n• Fix a : ℝ\n" without a closing Qed).
+        if (this.isBusy) {
+            return InputAreaStatus.Incorrect;
```
pimotte, line 181: This is kind of a sledgehammer approach: We don't actually consider part of the document, we just mark all areas as incorrect. We eventually want show partial progress: i.e. marking areas green as we can.

I can imagine fixing this is dependent on #319, can we add this onto that issue?

pimotte (commented):
I've done a manual test, it seems to work okay, but I had a crashing document as well. This might be fixed by the commit you just pushed.

If you explicitly re-request review when you've pushed everything you want to, I'll do a final round of testing.

GitLuckier:
Ok, I just spent a long time trying to hunt down a bug. The fix turned out to be just adding the following guard to check for a `null` value:

```typescript
        if (!response) {
            return InputAreaStatus.Incorrect;
        }
```

The reason it took me so long is because it doesn't give a type error when I remove this guard. It turns out that `leanGoalRequestType` inside `src/lsp-client/lean/requestTypes.ts` is typed wrong. It should be:

```typescript
export const leanGoalRequestType = new RequestType<LeanGoalRequest, LeanGoalAnswer | null, void>("$/lean/plainGoal");
```

Note the added `null`. The fact that it was missing made typescript think there was no null check needed.

Should I work on fixing this type issue next, after this PR has been reviewed?

pimotte:
I would just include the fix in this PR, it seems simple enough. (This is the kind of thing I meant with "doing cleanup on code you're working", even though you strictly don't have any edits for that). If it crosses to waterproof-editor, then you'll need an additional PR, but that's fine.

pimotte (approved):

PR #334, opened 2026-03-30
Sorry is invalid, hint is incorrect, isBusy flag properly resets.

### LINKED ISSUE #328: [Bug] [Lean] Use of hint and sorry marks input area as correct
Followup issue to #313, mentioned in #327 

The use of hint or sorry should not mark the input area as correct (i.e. mark it green).

Acceptance criteria:
- Any sorries should disqualify a proof from being green.
- help is only allowed to mark the proof green if the suggestion actually makes the proof correct, but also making use of help a disqualification for a correct proof is fine.

GitLuckier:
### Description
`sorry` and `hint` were marking the input areas as Correct. This should fix it. `sorry` is now marked as Invalid. The ` This pull request also fixes an issue where the `isBusy` flag would sometimes not properly reset.

### Changes
When a user types `sorry`, Lean will not mark the input as using `sorry`, but will mark the theorem. This means that the message "deceleration uses `sorry`" will appear outside the input area. In order to link the input area with this message, we look if there are any `sorry` messages between the end of the current input area and the end of the previous one. Then we mark that input area as invalid. Same for `hint`, since it also sends `sorry` messages when used.

I also removed a mistake I made with the `isBusy` flag being set with this function called `onDocumentChanged`. This function shouldn't be in there any more, so I removed it. This fixes a bug where correct input area fields would stay permanently red, because `isBusy` wasn't being set to false. 

renamed `client.determineProofStatus.test.ts` to `client.proofStatus.test.ts` to be more inclusive of functions other than determineProofStatus.

### Testing this PR
- Open your favourite Lean document.
- Type `sorry` or `hint`.
- Observe that only the input area you selected is turning yellow
- Input areas that are correct should no longer stay red forever when typing in other input fields. (`isBusy`)

There are also a few automated unit tests. Most notably, this approach requires `getInputAreas` to return a sorted array, I have written tests to ensure that this is always the case

Closes #328

### DIFF
--- a/__tests__/lsp-client/lean/client.proofStatus.test.ts
+++ b/__tests__/lsp-client/lean/client.proofStatus.test.ts
@@ -76,16 +76,7 @@ jest.mock("../../../src/helpers",   () => ({
 jest.mock("../sentenceManager",     () => ({ SentenceManager: class { onProgress() {} dispose() {} } }), { virtual: true });
 jest.mock("../clientTypes",         () => ({}), { virtual: true });
 jest.mock("../requestTypes",        () => ({ convertToSimple: jest.fn() }), { virtual: true });
-jest.mock("../qedStatus",           () => ({
-    findOccurrences: jest.fn((substr: string, str: string) => {
-        const indices: number[] = [];
-        const substrLen = substr.length;
-        for (let i = 0; (i = str.indexOf(substr, i)) >= 0; i += substrLen) {
-            indices.push(i);
-        }
-        return indices;
-    }),
-}), { virtual: true });
+
 jest.mock("./requestTypes",         () => ({
     leanFileProgressNotificationType: "$/lean/fileProgress",
     leanGoalRequestType:              "$/lean/plainGoal",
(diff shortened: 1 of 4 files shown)


__tests__/lsp-client/lean/client.proofStatus.test.ts
```diff
@@ -0,0 +1,590 @@
+jest.mock("vscode", () => {
```
pimotte, line 6: Seeing these mocks kind compared to the complexity of the code make me think that the tests are maybe at the wrong abstraction level? I have no objection to keeping them as integration tests if you think they're useful, but it might make sense to write these more as unit tests.

pimotte (changes requested):
I'm a bit confused about the scope of this PR. In my testing, the behaviour is:

- Correctly changes to yellow for sorry
- Stays red for help (is fine)
- Incorrectly changes to green for hint (incorrect, but also not the most important per the issue)

The PR description says we tackle sorry and hint, but I don't see any code for either hint or help.

Can you clarify/document what this is supposed to cover?

GitLuckier:
> I'm a bit confused about the scope of this PR. In my testing, the behaviour is:
> 
>     * Correctly changes to yellow for sorry
> 
>     * Stays red for help (is fine)
> 
>     * Incorrectly changes to green for hint (incorrect, but also not the most important per the issue)
> 
> 
> The PR description says we tackle sorry and hint, but I don't see any code for either hint or help.
> 
> Can you clarify/document what this is supposed to cover?

The `hint` is actually also covered in this PR. Fun fact I discovered about Lean: the `hint` keyword also produces a `sorry` (you can see this for yourself in any lean file). This is what caused the input area to be marked as correct when typing `hint`. Fixing `sorry` therefore fixes `hint` for free. (This also works for the `admit` keyword, which also produces a `sorry` message)

About the `help` keyword: I think the existing implementation is fine. If we still want to change its behaviour, it would be out of scope for this PR, as it is not really related to the `sorry` issue.

About testing: given how heavily mocked these tests already are, I'd argue they're closer to unit tests than integration tests. The mocks just happen to be verbose because of the VSCode API surface. Happy to discuss if you think there's a better way to structure them though.

pimotte:
I've looked again, and I think my cause for confusion is that hint does turn the proof green when it solves the goal.
image

I think this is fine, but I do think we should document the current behaviour at determineProofStatus at lean/client.ts. Other than that, I'm happy with the code as-is. I think the test is more of an indication we could refactor this a bit to be more modular in some way, but I think this is a) low priority and b) out of scope for this PR.

pimotte:
Thanks!

PR #336, opened 2026-03-30
fix: "verified up to" bar no longer gets stuck (minor refactor)

GitLuckier:
### Description
Fixed the bug where the progress bar would get stuck, even if lean was done checking. This time a minimal fix without the refactoring. 

The fix was to ensure the progress array is non empty when you try to access `progress[0]`:

```typescript
		case MessageType.progress:
			{
				const { numberOfLines, progress } = msg.body;
				if (progress.length === 0) {
					editor.removeBusyIndicators();
					break;
				}

				const at = progress[0].range.start.line + 1;
				if (at === numberOfLines) {
					editor.reportProgress(at, numberOfLines, "File verified");
				} else {
					editor.reportProgress(at, numberOfLines, `Verified file up to line: ${at}`);
				}
				break;
			}
```

Instead of the major refactoring I did earlier, I could extract the messaging logic into its own file.


pimotte (changes requested):
Yeah, I think that's actually what we really want! The logic in the listener is not too fancy, but clearly we can have bugs here, so I think it makes sense to extract the entire listener to a separate file and unit test them, in this PR.

GitLuckier:
I extracted all the messages send from waterproof-vscode to the editor into a file with a pure function. This should improve readability and allow us to write tests. I already wrote some tests for the non-trivial cases. This should catch bugs like this one in future.

pimotte (approved):
Looks good to me, tested locally, thanks!

PR #337, opened 2026-03-30
Fix: Save as exercise sheet

### LINKED ISSUE #314: [Bug] [Lean] Save as exercise-sheet does not empty input areas
### What happened?
With a Lean exercise sheet, it should be possible "Save as exercise sheet"

Example:
1. Open the Lean ATP file
2. Execute the "Save as exercise sheet" command: Ctrl + Shift + P > Waterproof: Save as exercise sheet
3. Open the saved file and notice that the input areas are not empty.

Save as exercise sheet should remove solutions to exercises by emptying input areas.

Implementation notes:

We want to take this opportunity to refactor some stuff as well. 

Plan:

- Update typescript implementation of this functionality to include Lean files
- Make sure typescript implementation is callable in script form
- Convert [this Python script](https://github.com/impermeable/waterproof-vscode/blob/main/scripts/export_exercise_sheet.py) to a typescript script that uses the implementation used for the command.
- Expose the compiled version of this typescript script to something that can be run with `npx`
- Update the CI in [this repo](https://github.com/impermeable/Bewijzen-in-de-Wiskunde) to use `npx @waterproof/vscode something something` to do the conversion, rather than the existing python script.

Michaillus:
### Description

- This adds saving as exercise sheet functionality for .lean files (in addition to rocq .mv) in vscode. Code inside such input block gets emptied:
````
:::input
```lean
[some code here]
```
:::
````
- This adds save as exercise sheet functionality for .lean files in a script runnable from github actions. The script is rewriten from python to typescript.

- There will be a coordinated pull requrest in Bewijzen-in-de-Wiskunde that changes the github action script.

Closes #314 

### Changes
- The function that removes code from input areas of .mv or .lean file is separated in a helper file. This function is used by both vscode save as exercise sheet action and typescript script.
- export_exercise_sheet.py python script is removed and substituted with typescript script performing the same logic.

### Testing this PR
- To test the vscode action:
Open waterproof vscode extension and perform Ctrl+Shift+P -> Save as exercise sheet action on a waterproof rocq and lean file. It should create a file where code in input areas is removed.
- To test the script:
It is tested through github action of Bewijzen-in-de-Wiskunde repository, I've done some testing through github action runs of refactor-export-action branch.
Also, it can be tested by clean cloning waterproof-vscode repo, running
````
npm install typescript @types/node ts-node commander
````
and then the script itself by
````
npx ts-node-esm --skipProject waterproof-vscode/scripts/exportExerciseSheet.ts .....
````
where the dots should be substituted by the target file or input+output folders.
This is also how it is done in github actions.

### CODE REVIEW

File: src/helpers/exerciseSheet.ts [hunk 1]
```diff
@@ -0,0 +1,43 @@
+/**
+ * Process Waterproof Rocq or Lean file content by removing solutions from input-area blocks.
+ * Finds input area blocks of the form and replaces the content with empty code blocks.
```
pimotte: Let's also explicitly document that this does not work for multiple code blocks in a single input cell. (This is fine for me, this is how we build sheets currently, but it's a bit of a trap.)

src/helpers/exerciseSheet.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * ....code....
+ *
+ * ```
+ * :::
+ *
+ * @param content - The raw content of a .mv or .lean file
+ * @returns Processed content with solutions removed
+ */
+export function processWaterproofContent(content: string, extension: string): string {
```
pimotte: Could you add a unit test for this function for the .mv and .lean case?

scripts/exportExerciseSheet.ts
```diff
@@ -0,0 +1,129 @@
+import * as path from 'path';
```
pimotte, line 33: Can you document the different ways of running this script here (a little more explicit than in the PR, with concrete examples.?) Then we can easily find it when trying to run this.

scripts/exportExerciseSheet.ts
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Get paths of all .mv and .lean files inside a folder recursively.
+ *
+ * @param folder - Path to the folder
+ */
+function getAllFiles(folder: string): string[] {
+    let allFilePaths: string[] = [];
+    const fileNames = fs.readdirSync(folder);
+
+    fileNames.forEach((fileName) => {
```
pimotte, line 53: If I might request a small extra feature, can we exclude ".lake" folders here? These are in the repositories with lean exercises, but we really don´t want to process them.

pimotte (changes requested):
Thanks! Core functionality works nicely:) I do have some small remarks for maintainability/ease of use.

src/helpers/exerciseSheet.ts
Michaillus: Didn't know two code blocks in one input cell were possible. Is this desirable in some cases? I think it won't be too hard to play around with regex and make it work.
I tried creating input cell with two code blocks with rocq like this
````
<input-area>
```coq
something
```
```coq
something else
```
<\input-area>
````
but it just show these two lines together in an input cell.

For now I'll document as you asked.

Michaillus:
Added all the requested changes

src/helpers/exerciseSheet.ts
pimotte: It is not desirable per se, but the editor does allow it, and you could do stuff with it like: Have a code cell, then a markdown area where a student writes some reflection, then another code cell.

pimotte (approved):

PR #329, opened 2026-03-18
Multilean

pimotte:
This PR is largely the product of Claude. I have reviewed the non-testing code in detail and am happy with it. I have skimmed the tests. This PR has a [companion PR](https://github.com/impermeable/waterproof-editor/pull/82) in waterproof-editor.

### Description
Uses the generic container functionality to provide multilean support.

### Changes
- Adds multilean support
- Fixes esbuild to support npm workspaces setup

### Testing this PR

See: https://github.com/impermeable/waterproof-editor/pull/82

PR #340, opened 2026-04-02
Containers + bugfixes

pimotte:
This contains the changes from #329 plus a small bugfix in the Lean parser.

This is the companion PR to https://github.com/impermeable/waterproof-editor/pull/87


editor/src/index.ts
```diff
@@ -77,6 +77,17 @@ function createConfiguration(format: FileFormat, codeAPI: VSCodeAPI) {
 					highlightLight: langVerbose.highlight_light,
 					languageSupport: langVerbose.verbose(),
 				},
+				menubarEntries: [
+					{
+						title: "M...",
```
Tammo0987, line 83: what does "M..." mean?

Tammo0987 (changes requested):

editor/src/index.ts
pimotte: Multilean, but that's too long to fit on the button xD

The mouseover on the button explains what it does, but I agree this is not ideal.
If we want to redesign the buttons, I propose to split that off to a separate issue.

jim-portegies:
@pimotte I seem to be having trouble here with the "Adopt suggestion" buttons.

pimotte:
> @pimotte I seem to be having trouble here with the "Adopt suggestion" buttons.

I have reproduced this on main and created #349. If "clicking quickly" does not resolve your issue, can you give some more details?


Unless there's more than this, I propose to do the fix outside of this branch.

Issue #136, opened 2025-01-13
[Bug] Title There is a problem with saving and the setting "trim trailing white spaces on save" in vscode.


jim-portegies:
### What happened?
Desynchronization of the editor can happen with the setting vscode "trim trailing whitespaces on save" on, leading to broken files.

### How can we reproduce this issue?

1. Turn on the setting "Trim Trailing Whitespace" in vscode.
2. Open the tutorial.
3. Go to the first input area.
4. Enter two spaces (or for a more dramatic effect ten spaces).
5. Save the document with Ctrl+S
6. Press backspace
7. The document breaks, the closing tag for the code block only has two backticks left.

### What was supposed to happen?

The code block remains intact

The reason seems to be that the editor and the underlying file are out of sync after saving, so the position counting is off after saving.

pimotte:
Solution: Resolve this by warning the user if trim whitespace on save is on, and recommend turning it off in the workspace settings.

pimotte:
Fix is in #171, but needs to be ported to mapping branch

pimotte:
Resolved with above PRs

PR #345, opened 2026-04-13
Fix for trimmed whitespace issue 

raulTUe:
Description
This is a pull request to partially solve the trimming whitespace Issue #136. We now reload the editor only if the file has changed on save, and always give a warning to the user when the trim trailing whitespace setting is on.

Changes
The editor effectively re-initializes if the underlying file has changed when the file was saved, which only really occurs when the trim trailing whitespace setting is on. This makes it clunky if auto save is on, but the warning should prevent this.

Testing this PR
When initializing the extension, the warning should appear on the bottom left of the screen. Moreover, if you turn on trim trailing whitespace setting, and then add a couple of spaces at the end of one of the input areas, and press ctrl+s, then the editor will  refresh the text in the editor with the content at disk (i.e no whitespace) while also staying in the same position in the document.


developer-resources/messages.md
```diff
@@ -123,6 +123,19 @@ Send by the extension to start initialization of the webview editor, includes th
 }
 ```
 
+### `refreshDocument`
+
+#### Description
+Send by the extension to refresh the webview editor, includes the updated content of the document in the body, as well as the documnet's version.
```
DikieDick:
```suggestion
Send by the extension to refresh the webview editor, includes the updated content of the document in the body, as well as the document's version.
```

shared/Messages.ts
```diff
@@ -71,6 +72,7 @@ export const enum MessageType {
     executionInfo,
     flash,
     init,
+    refreshDocument,
```
DikieDick: Not a deal breaker for merging this PR, but this added line is one position away from keeping the list of types alphabetical :)

src/util.ts
```diff
@@ -53,3 +53,23 @@ export function excludeCoqFileTypes() {
...
+/**
+ * Checks whether the user has enabled the setting `trimTrailingWhitespace`, and warns the
+ * user in that case.
+ */
+export function checkTrimmingWhitespace() {
+    const config = workspace.getConfiguration('files');
+    const isTrimTrailingWhitespaceEnabled = config.get<boolean>('trimTrailingWhitespace');
+    if (isTrimTrailingWhitespaceEnabled) {
+        window.showWarningMessage(
+            "The setting `Trim Trailing Whitespace` is enabled. This may cause unexpected behaviour, and is strongly suggested to turned off.",
```
DikieDick: The end of this sentence is not really correct, I think. 

```suggestion
            "The setting `Trim Trailing Whitespace` is enabled. This may cause unexpected behaviour, and is strongly suggested to be turned off.",
```

Alternatively, we could say something like:

```suggestion
            "The setting `Trim Trailing Whitespace` is enabled. This may cause unexpected behaviour in Waterproof, and we thus recommend you to turn it off.",
```

src/util.ts
```diff
@@ -53,3 +53,23 @@ export function excludeCoqFileTypes() {
...
+            "Open Settings",
+            "Dismiss"
+        ).then((selection: string | undefined) => {
+            if (selection === "Open Settings") {
+                commands.executeCommand('workbench.action.openSettings', 'Trim Trailing Whitespace');
```
DikieDick, line 71: Very nice

DikieDick (changes requested):

pimotte (approved):
Thanks!

PR #346, opened 2026-04-13
Fix/edit after #doc

### LINKED ISSUE #321: [Bug] [Lean] Editing after #doc in technical details ends up in the wrong block
### What happened?
When editing a Lean-based exercise file, edits after #doc end up going to markdown instead of codecells.

Example:
1. Open any Lean-based exercise sheet
2. Unfold the Technical details section at the top
3. Set the cursor at the end of the doc line
4. Type enter and text
5. Close and open the document.
6. Find that the text is in a markdown cell.

Either the text should end in a code cell or the edit should not be allowed in the first place.

I think this is kind of a niche occurrence and teachers likely won't be doing this to begin with. I'm good with either of these solutions, but not allowing the edit seems more robust. If we allow the edit, it will still be hard to keep the document consistent between editor and disk.

Michaillus:
### Description
Fixes the bug - in lean waterproof files if you go at the end of the preamble, "#doc ... =>" line, and write anything after, it would move to a separate markdown block.

### Changes

I removed the "#doc ... =>" line from the visible range of the preamble CodeBlock, by reducing the inner range.
While I was working on the issue and familiarizing to the code, I also cleaned up some things.
 - Removed redundant ? from Kind.NewLine regex. It maches one newline either way.
 - Added some comments for editor provider, prose mirror webview and the tokenization with splitting in blocks.
 - In pmWebview, vscode WebviewPanel was called webview, which is confusing, so I renamed it to webviewPanel. Webview is our prose mirror webview that is displayed inside that panel.

### Testing this PR

Open a Waterproof Lean file and expand "technical details" preamble at the top. It should not contain "#doc .... =>" line at the end.
Switch to teacher mode. Click at the end of preamble block. Press enter and try to type something. Reopen the file and check that the text stays inside preamble block.

Closes #321


DikieDick (commented):
Nice fix!

Preliminary remark: Can we unit test this functionality?

Michaillus:
I added a test.
Also, I spotted that prosedoc-construction/vFile.test.ts test file is failing. It is failing in main as well, currently.

__tests__/lean-split-in-blocks/lean-split-in-blocks.test.ts
```diff
@@ -0,0 +1,30 @@
...
+∀ ε > 0, ∃ N, ∀ n ≥ N, |u n - l| ≤ ε
+
+def continuous_function_at (f : ℝ → ℝ) (x₀ : ℝ) :=
+∀ ε > 0, ∃ δ > 0, ∀ x, |x - x₀| ≤ δ → |f x - f x₀| ≤ ε
+
+notation:50 f:80 " is continuous at " x₀ => continuous_function_at f x₀
+notation:50 u:80 " converges to " l => sequence_tendsto u l
+`;
+
+    const inputPath = path.join(__dirname, 'input.lean');
```
pimotte: I'd like to suggest two small changes to this test:

- Embed the string in the test file so we don't have to go through fs/path. (For consistency with other tests, and to make it self-contained.)
- Also test the second block

pimotte (approved):
Works well. I do have two small nitpicks on the test, but feel free to merge after addressing those.

pimotte:
> I added a test. Also, I spotted that prosedoc-construction/vFile.test.ts test file is failing. It is failing in main as well, currently.

This is fixed in #340

Michaillus:
I applied the suggestions

pimotte (approved):
Thanks!

PR #348, opened 2026-04-14
fix: accidentily copied the wrong switch statement into messageHandler.ts

GitLuckier:
I messed up :(  In PR #336, I accidentally copied the buggy version into messageHandler.ts, which meant the file progress bar still got stuck. I replaced it with the fixed version and updated the tests.


pimotte:
Mmmmm. I'm kind of confused what happened then, because I did test the original PR locally. I have tested this and observed that this indeed fixes the issue compared to mapping.

PR #351, opened 2026-04-27
Add pos argument to tryStep and return range of proof via proofContext

DikieDick:
- Bumps version to 0.11.1 (which is odd, so prerelease?)
- Adds the option of trying a proof step at a given position, by updating the API and executeCommandBase.
- proofContext now returns the range of the proof script (between `Proof.` and `Qed.`, but works for scripts without `Proof.` as well, see the regex)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "waterproof",
   "displayName": "Waterproof",
   "description": "Waterproof for VSC",
-  "version": "0.10.1",
+  "version": "0.11.1",
   "waterproofConfiguration": {
     "requiredCoqLspVersion": ">=0.2.4",
     "requiredCoqWaterproofVersion": ">=3.1.0",
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,4 +1,4 @@
-import { Position, TextDocument } from "vscode";
+import { Position, Range, TextDocument } from "vscode";
 import { RunResult } from "./lsp-client/petanque";
 import { GoalConfig } from "../lib/types";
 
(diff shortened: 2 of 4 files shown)


pimotte (commented):
I think the code looks fine, but I do have some questions:

- Just to check, this PR targets main intentionally?
- Was there a way I could easily test these API methods interactively?
- Does it make sense to unit test the commandExecutor?
- Do the proofContext changes have an implication for the Lean functionality when backmerging this into mapping?

DikieDick:
> Just to check, this PR targets main intentionally?

Yes, as River currently works best (maybe even 'only') on main.

> Was there a way I could easily test these API methods interactively?

Apart from using River's `translateProof` with a faulty proof and cursor not at the start of the input area, no.

The main problem this solves is that River would otherwise only be able to start a petanque context at the cursor position, instead of a blank input area.

> Does it make sense to unit test the commandExecutor?

Yes, but we can't fully test it without having a valid LSP in the background.

> Do the proofContext changes have an implication for the Lean functionality when backmerging this into mapping?

To my knowledge all this API stuff is only ever used for River, which relies on Rocq

pimotte:
I think it does make sense to unit test the executor with a mocked client. While it's not much code, I think it's good to err on the side of testing a bit too much, and it also documents the assumptions we make on the LSP output for purposes of the executor. Especially if it's easier to setup, a test with the real client but mocked LSP messages would also be a good option.

DikieDick:
Superseded by #395

PR #353, opened 2026-05-06
feat: proposal on how to resolve scary warning

### LINKED ISSUE #332: [Bug] [Lean] Starting a subproof gives a scary warning
When starting a subproof with \. the unproven goals warning is not located on the QED, but on the subproof dot. This can look kind of scary, because the warning show the whole state:

Image

Solving this requires some thought about what the most maintainable solution is, it might be more on the Lean side, rather than on the frontend side.

GitLuckier:
### Changes
The Lean LSP inlcudes non-standard fields inside their diagnostic messages. These properties get lost when vscode converts them. The Lean4 vscode extension rescues those fields by monkey patching the converters:

See: https://github.com/leanprover/vscode-lean4/blob/bbe3ff126cd54defd10ac6e07a2b94dd2d865c42/vscode-lean4/src/utils/converters.ts

This PR uses a more minimal patch in order to rescue just those leanTags and some other fields. This also solves an issue where we had these existing types in `waterproof-vscode/src/lsp-client/lean/requestTypes.ts`:

```ts
`export interface LeanDiagnostic extends Diagnostic {
    fullRange?: Range;
    isSilent?: boolean;
    leanTags?: LeanTag[];
}
```

but they were never included in the final diagnostics message. Now that they are, the type makes sense.

It then uses those leantags to determine whether a diagnostics message is about `unsolved goals`. This means we have a flag to determine whether we should hide this `unsolved goals` error message in the editor (see #332 ). This is more robust then string matching on `unsolved goals` in the diagnostic. Furthermore, this patching allows us to add more patching similar to the Lean4 vscode extension if needed in the future.

### Testing this PR
If you really want, you could inspect the new contents of the diagnostics messages to see the leanTags now being preserved. Otherwise there are some new unit tests.


GitLuckier:
I have a new proposal that replaces the scary messages with a smaller (hopefully more helpful) one. I could have changed the diagnostic severity of the message. However, after testing, I came to the conclusion that keeping at an `Error` level of severity made the most sense. This is because making it an `Information` would have been too subtle of a squiggle and making it a warning actually felt more intimidating to me than an error.

Let me know what you guys think of this proposal.

src/lsp-client/lean/converter.ts
```diff
@@ -0,0 +1,40 @@
+import { Code2ProtocolConverter, Protocol2CodeConverter } from "vscode-languageclient";
+import { LeanDiagnostic } from "./requestTypes";
+import { Diagnostic } from "vscode";
+
+import * as async from 'vscode-languageclient/lib/common/utils/async'
+
+/**
+* Preserve Lean-specific diagnostic fields (e.g. leanTags) during protocol<->code conversion.
+* Vscode does not preserve them on its own.
+* Inspired by the lean4 vscode extension's approach:
```
pimotte, line 11: Let's include a permalink to the lean4 extension code here

src/lsp-client/lean/client.ts
```diff
@@ -25,21 +26,25 @@ export class LeanLspClient extends LspClient<LeanGoalRequest, LeanGoalAnswer> {
...
         this.disposables.push(this.client.onNotification(leanFileProgressNotificationType, params => {
             this.onFileProgress(params);
         }));
 
         const hndl = this.client.middleware.handleDiagnostics;
         this.client.middleware.handleDiagnostics = (uri, diagnostics_, next) => {
-            if (hndl) hndl(uri, diagnostics_, next);
+                
+            const inputAreas = this.activeDocument ? this.getInputAreas(this.activeDocument) : undefined;
+            const friendlyDiagnostics = this.rewriteDiagnostics(diagnostics_, inputAreas);
```
pimotte: Should we be rewriting diagnostics if the input areas are undefined?

src/lsp-client/lean/client.ts
```diff
@@ -50,6 +55,27 @@ export class LeanLspClient extends LspClient<LeanGoalRequest, LeanGoalAnswer> {
...
+     * @returns The rewritten diagnostics.
+     */
+    private rewriteDiagnostics(diagnostics: WpDiagnostic[], inputAreas: Range[] | undefined): WpDiagnostic[] {
+        return diagnostics.map(d => {
+            if ((d as LeanDiagnostic).leanTags?.includes(LeanTag.UnsolvedGoals)) {
+                const isInsideInputArea = inputAreas?.some(area =>
+                    area.intersection(d.range) !== undefined && !area.intersection(d.range)!.isEmpty
+                );
+                if (isInsideInputArea) {
+                    return { ...d, message: `Subproof starting on line ${d.range.start.line + 1} has unsolved goals.` };
```
pimotte: I'm tempted to make the language "is not finished yet." rather than "has unsolved goals.", this seems a bit friendlier.

pimotte (changes requested):
Looks good to me, tested and is a nice improvement to the status quo.

I do have some small nitpicks, after I think this is ready for merge.

pimotte (approved):
Works well, LGTM, thanks!

PR #377, opened 2026-06-17
Fix stale positioning bugging infoview update

### LINKED ISSUE #371: [Bug] [Lean] Infoview does not update with error until cursor moves
### What happened?
When typing a non-existent tactic, the error only shows in the infoview after moving the cursor.

Example:
1. Open a lean exercise sheet
2. Type "blah" in an input area
3. Observe no error
4. Move cursor 
5. Observe error

The infoview should be updated sooner

pimotte:
### Description
Fixes #371 

### Changes
- Ensures position is requested at the correct time.
- Adds tests

### Testing this PR
- Look at #371 and confirm this behaviour has been fixed.


pimotte:
Audit is fixed in #378

PR #380, opened 2026-06-25
Fix issue that the info view gets stuck

### LINKED ISSUE #374: [Bug] [Lean] Clicking into the file quickly breaks the proof view
Example:
1. Start the extension with a Lean file open
2. Immediately click a proof area.
3. Observe the infoview being stuck at `Waiting for Lean server to start...`

It should update to show the goals.

(Probably some sort of race condition)

Tammo0987:
### Description
This PR fixes a race condition to get stuck in waiting for the lean server. The fix registers a one time handler for the event, that the lsp is running and does the same action in the happy path.

### Testing this PR
Test manually, but also because of timing hard to reproduce.

Fixes #374

### DIFF
--- a/src/infoview.ts
+++ b/src/infoview.ts
@@ -9,7 +9,6 @@ import {
 } from "@leanprover/infoview-api";
 import { Rpc } from "./helpers/rpc";
 import { MessageType } from "../shared";
-import { State } from "vscode-languageclient";
 import {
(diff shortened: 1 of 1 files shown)


pimotte (approved):
Tested, though my computer seems such it was hard to reproduce the original issue cleanly, so I think we're good in any case:P

PR #381, opened 2026-06-25
Fix adoption bug in rocq by removing insert support for goal panel

### LINKED ISSUE #349: [Bug] Adopt suggestion button is extremely finicky
Example:
1. Open a .mv file
2. Produce a situation with an "Adopt suggestion button" (e.g. type Either a ≥ 0 or a ≤ 0. in a proof)
3. Click the adopt suggestion button
4. Observe it only insert the suggestions about 10-50% of the time, depending on how you click.

Alternatively: Click quickly and see it succeed a lot more.

It should insert the suggestion every time. 

Hypothesis: Clicking for a longer time or slightly moving the mouse while clicking moves the focus to the goals panel, removing the cursor from the editor, and thus failing to obtain a place for the suggestion.

Tammo0987:
### Description
This PR fixes a bug which seems to be accidentally introduced within a refactor by enabling insert support for the goal panel.

### Changes
- Fix the bug
- Removed unused code.

### Testing this PR
Use a rocq mv file and test it manually as stated in the issue #349.

### DIFF
--- a/src/webviews/goalviews/goalsPanel.ts
+++ b/src/webviews/goalviews/goalsPanel.ts
@@ -4,16 +4,14 @@ import { WebviewEvents, WebviewState } from "../waterproofPanel";
 import { MessageType } from "../../../shared";
 import { GoalsBase } from "./goalsBase";
 import { IExecutor } from "../../components";
-import { PpString, RocqGoalAnswer } from "../../../lib/types";
 
 //the goals panel extends the GoalsBase class
 export class GoalsPanel extends GoalsBase implements IExecutor {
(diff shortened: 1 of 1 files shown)


pimotte (approved):
I'm kind of baffled by this, but it works:)

If you have any insight into why this caused the inconsistent behaviour, could you add this in the PR description or in a sensible place in the code?

Feel free to merge, even if you don't have the explanation.

Tammo0987:
So I think when we have true we have two interfering panels over each other where the click events then get the race condition. It was actually never intended to have true for this panel, but it got changed to true during solving a merge conflict. At least it looks like that in history, so now it kind of returns to false as the original value (by default).

PR #383, opened 2026-07-01
Add prelimary Yalep support

pimotte:
Turns on when a file ends with _yalep.lean.

Uses highlighting from Lean, so that might be broken.


jim-portegies (approved):
This looks mostly good, and I think we should merge, but I do think eventually we would like to find a different way to find out that something is a yalep file.

PR #395, opened 2026-07-14
Add pos argument to tryStep and return proofRange via proofContext

DikieDick:
- The optional pos argument can be used to execute a proof step at that pos in the document.
- The proof range that is returned is the range of the user written proof in the document.

Both are required for `translateProof` to work in the latest version of River.

### DIFF
--- a/__tests__/extension/proofContext.test.ts
+++ b/__tests__/extension/proofContext.test.ts
@@ -169,11 +169,7 @@ type Ctx = {
   };
 };
 
-function makeCtx(
-  document: unknown,
-  cursor: unknown,
-  symbols: Sym[],
-): Ctx {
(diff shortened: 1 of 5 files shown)


DikieDick:
@pimotte Created this PR to supersede #351 as this was quicker. Can you take a look at unit testing the newly added functionality?

pimotte (approved):
In the process of adding the unit tests, I (read: Claude) found a bug with the regex.

This PR also does not include the version bump, but we'll do that when we actually make the release.

@DikieDick Could you check that https://github.com/impermeable/waterproof-vscode/pull/395/changes/10a61bed584b4c22c4be72d4e6ace6180d761c4e is the correct fix? If yes, feel free to merge.

DikieDick:
@pimotte The fix for the regex looks perfect. 

I have made a little change to also work with `Goal` and `Example`.

PR #407, opened 2026-08-13
Feat code actions verbose lean

### LINKED ISSUE #372: [Spike] [Lean] Broaden special casing of suggestions?
Currently, when calling help, we show the message, but this is not interactable.

Image

We should likely extend the special handling of messages from Rocq-waterproof to have the replace/insert buttons here too.

Targets of this spike:
- Figure out if this is doable on the front-end side
- Determine if it is a hard requirement or would be nice to have, to have Verbose Lean emit separate messages.

GitLuckier:
Companion PR:
https://github.com/impermeable/waterproof-editor/pull/123

### Description
Adds support for code actions for any diagnostic to be sent to the editor, in particular for Verbose Lean's `help` tactic.

### Changes
- Adds the ability to retrieve code actions from any arbitrary LSP (not just the Lean LSP).
- Sends those actions lazily to the editor (i.e., it will first send the diagnostics, then the corresponding actions, because some actions can take a while. This way we don't block diagnostics while waiting for the code actions).
- Adds a cancellation system to throw away actions when the document has changed.
- If an action is applied outside the input area, we don't show it to the user (e.g., "import this library" actions).
- Each LSP can control which actions to show. This way, we can tightly adjust which actions are shown to the user. For the Lean LSP, we show actions from the "Try this" provider, which are actions from the hint/help-like tactics.
- I have written some AI-assisted unit tests, though they have been verified by hand to properly catch real-world regressions. 
- We limit ourselves to 3 actions (a bit arbitrary, but should be more than enough for the `help` tactic).

### Testing this PR
- When typing in `help`, you should be able to see buttons appearing saying "Try this: ". If you click on those, the `help` tactic is replaced by the corresponding suggestion.
- Retrieving these actions is quite slow in some cases, especially from the Lean LSP (some actions can take several seconds). Since the LSP requests are cancellable on document changes and are retrieved after the diagnostics themselves have already been sent, the user should not notice anything regarding the slowness. Please notify me if this is not the case. 
- It should not show actions that shouldn't be visible to the Waterproof user (like "import this library").
- Actions should persist between tab switches. 
- There are unit tests. 

While this PR technically does not require any changes to Verbose Lean in order to work, I do have one suggestion. For the `hint` tactic, Lean itself adds the suggestion to the title of the action, so you get `Try this: {actual suggested text}`:

image

The verbose `help` tactic just sends "Try this:", which in practice can look like this:
image

Notice that the buttons don't show which code is actually applied. To fix this, we can reuse Lean4's `hint` code in Verbose Lean as follows:
```lean
def mkSuggestionsMessage (suggestions : Array Suggestion) (ref : Syntax) : CoreM MessageData := do
  let mut msg := m!""
  for suggestion in suggestions do
    let some range := ref |>.getRange? | continue
    let edit ← suggestion.processEdit range
    let suggestionText := edit.newText
    let ref := Syntax.ofRange <| ref.getRange?.getD range
    let info := Info.ofCustomInfo {
      stx := ref
      value := Dynamic.mk {
        edit
        suggestion := suggestion
        codeActionTitle := (← try_this)
        : Lean.Meta.Tactic.TryThis.TryThisInfo
      }
    }
```
to
```lean
def mkSuggestionsMessage (suggestions : Array Suggestion) (ref : Syntax) : CoreM MessageData := do
  let mut msg := m!""
  for suggestion in suggestions do
    let some range := ref |>.getRange? | continue
    let edit ← suggestion.processEdit range
    let suggestionText := edit.newText
    let ref := Syntax.ofRange <| ref.getRange?.getD range
    let info := Info.ofCustomInfo {
      stx := ref
      value := Dynamic.mk {
        edit
        suggestion := suggestion
        codeActionTitle := (← try_this) ++ suggestionText -- add suggestion to the title
        : Lean.Meta.Tactic.TryThis.TryThisInfo
      }
    }
```
Like I said, it isn't strictly necessary, but it would make me sad otherwise :(

Note: I used AI for initial prototyping.

Closes #372


src/lsp-client/client.ts
```diff
@@ -252,24 +259,236 @@ export abstract class LspClient<
...
+          );
+          continue;
+        }
+
+        validActions.push({
+          action: {
+            title: result.title,
+            edits,
+          },
+          isPreferred: "isPreferred" in result ? !!result.isPreferred : false,
```
pimotte: What's this !! ?

pimotte (commented):
Thanks! I think this generally looks good, but I'm not convinced on the proposed Verbose Lean change. I tried this locally:

Image

The Verbose Lean suggestions tend to be a lot longer than regular tactics. Furthermore, the button duplicates with the message. 

I think some alternatives might be:

- Monkeypatching the message to say "Try nth suggestion"
- Splitting the help message on the Verbose side to have the buttons tied to the message so it's clear 

What do you think?

src/lsp-client/client.ts
GitLuckier: For your first point, I agree. I have changed the styling slightly to now wrap around properly, which I think looks noticeably better: 
image
(Note: if one were to accept this PR right now, this is what would be merged)

 If that is still too much text. We could always try ellipses:
image

We can even make it expand to the side when you hover over it:
image

I personally would go for the hover-expansion approach, though feedback is always welcome!

If none of these suggestions floats your boat, out of your suggestions I'd gravitate towards option 1 (monkeypatching). I tried to implement number 2 (splitting the message), but I couldn't get far. I suspect it isn't possible without some major changes to either Verbose Lean or Lean itself, since this messaging is tied to Lean's implementation of the `TryThis` system, meaning that we either need to implement our own diagnostic messaging system or somehow alter Lean's implementation of it. Furthermore, if the input area contains both help messages and other diagnostics, it might not be clear to the user that the help messages are related. Monkeypatching requires some mental overhead for the user to relate the numbers to the actual message, but it creates the least amount of text. Another advantage would be that the student would be more likely to read the text associated with the `help` message.

For your second point, I apologise. I used AI to prototype the implementation and this is a leftover that slipped past my review. It is there because `validAction` has an `isPreferred` property that is strictly typed as a boolean. The `!!` is there to convert an `undefined` value into a boolean `false`. This makes sense to some degree, but the way `isPreferred` is used, it is redundant. I changed it to be simpler.

I also realized I forgot to note my LLM usage for prototyping in the original PR description, so I have updated it. Thanks for pointing this out!

pimotte: I agree the hover approach is very nice! Let's go with that.

GitLuckier: image

I tried implementing the hover expansion, but it requires the buttons to have a fixed width, which I don’t think looks as good in practice. I attempted to get the best of both worlds, but it is not possible using CodeMirror's styling. I’ve therefore gone with the simpler ellipsis version without hover expansion.

image

This makes the "📋" button on its own (without any additional action buttons) look like it always has:

image

Issue #364, opened 2026-05-28
[Spike] Possible memory leak in webextension


pimotte:
Based on some of the research in Utrecht I suspect the webextension might have a memory leak. Students describe the software becoming very sluggish. Specifically, this feedback makes me think this: "After using the software for more than 15 minutes the typing suggestions load very slowly. So when you want to type a special character using \, you have to type \, then wait a few seconds for the menu to load, and then type the character (if you start typing before the popup is loaded the typed special characters stays in text form and is not recognised by the software)."

Goal:

- Establish if there is a memory leak, or if this is just the issue of computation for the later exercise sheets.
- Check this for the Webextension in vscode.dev (working with the production version is fine), as well as the aquarium.

PR #411, opened 2026-08-31
Fix webextension

pimotte:
Fixes the webextension being broken (reported by Bart van den Dries) and probably the webextension having a memory leak (#364).

The webextension being broken was due the way the Lean and Rocq code paths were working in relation to isWeb. This has been slightly refactored.

Testing:

1. Run with the run task "Run as Webextension" and check that this works with a .mv file.
2. Spot check if this does not break the installed Rocq/Lean use

I have generated some tests to get a reasonable coverage on the changes, the remaining uncovered code is not really suitable for tests, in my opinion.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -41,3 +41,4 @@ completions/
 tsconfig.jest.json
 .claude
 hooks/
+coverage/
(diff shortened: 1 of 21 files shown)


jim-portegies (approved):
I have looked at this briefly and discussed with @pimotte , it looks good to me

